PR #3, opened 2016-08-23
DM-7344: Modernize & clean tests

timj:
The unittests in `meas_extensions_shapeHSM` have been modernized to make
them runnable under the py.test framework. They have been autoformatted
by autopep8 to be compliant with the LSST PEP8 standard and have been
checked for deprecated usages of assert_, assert<equals etc...

### DIFF
--- a/tests/testHsm.py
+++ b/tests/testHsm.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
+#
 # LSST Data Management System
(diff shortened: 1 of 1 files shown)


parejkoj:
It would be good to separate the autopep8 cleanups from the py.test cleanups, since those are separate things and there's a lot more of the former than the latter.

parejkoj:
While you're at it, please remove the `#-=-=` blocks.

PR #7, opened 2017-02-14
Add missing measure and fail methods

pschella:


r-owen (approved):
Looks fine.

It's surprising that "fail" was missing from the C++ code and that `hmsShapeControl.cc` had that copy/paste error (using the wrong class) and had no wrapping for `measure`. These suggest incomplete test coverage; could you please report that to @TallJimbo?

TallJimbo:
I've already got `shapeHSM` in my "needs better test coverage" list, but I'll add those concerns specifically to my notes.

PR #13, opened 2017-04-13
Tickets/dm 8108

natelust:


fred3m (commented):
Everything looks good, other than my preference for updating ``Footprint`` so that it can be initialized with a ``Point2I`` and ``width``, which would create a new ``SpanSet``from the ``Point2I`` and ``width``.

I don't see a reason to pass this on to the end user.

PR #15, opened 2017-09-22
Add PSF shape slot definition

SimonKrughoff:


PaulPrice (approved):
Thanks for including this.

PR #16, opened 2017-12-06
DM-12540: Add HSM moments that use a circular weight function

PaulPrice:


include/lsst/meas/extensions/shapeHSM/HsmAdapter.h
```diff
@@ -37,7 +37,7 @@ class ImageConverter
         // in the box and stride.
         PixelT* ptr = reinterpret_cast<PixelT*>(array.getData() + (_box.getMinY() - _image->getY0())*stride +
                                                 _box.getMinX() - _image->getX0());
-        return galsim::ImageView<PixelT>(ptr, _owner, stride, bounds);
+        return galsim::ImageView<PixelT>(ptr, _owner, 1, stride, bounds);
```
PaulPrice, line 40: This is great, but presumably this has to be linked with an update to GalSim, which I don't see on this ticket.

config/enable.py
```diff
@@ -6,7 +6,7 @@
 
 import lsst.meas.extensions.shapeHSM
 config.plugins.names |= ["ext_shapeHSM_HsmShapeRegauss", "ext_shapeHSM_HsmSourceMoments",
-                         "ext_shapeHSM_HsmPsfMoments"]
+                         "ext_shapeHSM_HsmPsfMoments", "ext_shapeHSM_HsmRoundMoments",]
```
PaulPrice: I'm not a fan of the naming choice: `ext_shapeHSM_HsmRoundMoments` in comparison with `...HsmSourceMoments` and `...HsmPsfMoments` suggests it's measuring the moments of a "round" (whatever that is). How about `...HsmSourceRoundMoments` or `...HsmSourceMomentsRound`?

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -40,17 +40,34 @@ class HsmMomentsControl {
 class HsmSourceMomentsControl {
 public:
     LSST_CONTROL_FIELD(badMaskPlanes, std::vector<std::string>, "Mask planes used to reject bad pixels.");
+    LSST_CONTROL_FIELD(roundMoments, bool, "Use round weight function.");
+    LSST_CONTROL_FIELD(addFlux, bool, "Store measured flux.");
```
PaulPrice: I think that if you're going to call this a fundamentally different source measurement algorithm, the `roundMoments` setting shouldn't be in the `HsmSourceMomentsControl` (which leaks out to the user) but in `HsmMomentsAlgorithm`.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -40,17 +40,34 @@ class HsmMomentsControl {
...
 
     HsmSourceMomentsControl(HsmSourceMomentsControl const & other) :
-        badMaskPlanes(other.badMaskPlanes) {}
+        badMaskPlanes(other.badMaskPlanes),roundMoments(other.roundMoments),addFlux(other.addFlux) {}
```
PaulPrice: Spaces after commas.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
PaulPrice: I think the help strings should end with question marks rather than periods.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -123,7 +143,12 @@ class HsmSourceMomentsAlgorithm : public HsmMomentsAlgorithm {
 
     /// @brief Initialize with standard field names and customized documentation.
     HsmSourceMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
-        HsmMomentsAlgorithm(name, schema, "Source adaptive moments algorithm from HSM"), _ctrl(ctrl) {}
+        HsmMomentsAlgorithm(name, schema, "Source adaptive moments algorithm from HSM"), _ctrl(ctrl)
+    {
+        if (ctrl.addFlux) {
+            _fluxKey = schema.addField<float>(name+"_Flux", "HSM flux");
```
PaulPrice: Space either side of `+` operator.

python/lsst/meas/extensions/shapeHSM/hsmMomentsControl.cc
```diff
@@ -45,6 +45,7 @@ PYBIND11_PLUGIN(hsmMomentsControl) {
     py::class_<HsmSourceMomentsAlgorithm, std::shared_ptr<HsmSourceMomentsAlgorithm>, HsmMomentsAlgorithm>
             clsHsmSourceMomentsAlgorithm(mod, "HsmSourceMomentsAlgorithm");
     py::class_<HsmSourceMomentsControl> clsHsmSourceMomentsControl(mod, "HsmSourceMomentsControl");
+    py::class_<HsmRoundMomentsControl, HsmSourceMomentsControl> clsHsmRoundMomentsControl(mod, "HsmRoundMomentsControl");
```
PaulPrice: Line exceeds length limit.

src/HsmMoments.cc
```diff
@@ -86,6 +88,12 @@ void HsmMomentsAlgorithm::calculate(
     base::ShapeResult shapeResult;
     shapeResult.setShape(Ellipse(ellip, radius));
     source.set(_momentsKey, shapeResult);
+    if (addFlux) {
+        if (!_fluxKey.isValid()) {
+            throw LSST_EXCEPT(pex::exceptions::LogicError, "Flux Key not defined");
```
PaulPrice: Lower-case "key".

Should this be an `assert` instead of an exception? You explicitly created the key earlier, so if it doesn't exist then it's a programming error.

src/HsmMoments.cc
```diff
@@ -125,8 +133,8 @@ void HsmSourceMomentsAlgorithm::measure(
     double const psfSigma = exposure.getPsf()->computeShape(center).getTraceRadius();
 
     HsmMomentsAlgorithm::calculate(source, exposure.getMaskedImage().getImage(),
-                          exposure.getMaskedImage().getMask(),
-                          bbox, center, badPixelMask, 2.5*psfSigma);
+                                   exposure.getMaskedImage().getMask(),
+                                   bbox, center, badPixelMask, 2.5*psfSigma, _ctrl.roundMoments, _ctrl.addFlux);
```
PaulPrice: I think this line is a tad longer than the limit.

tests/testHsm.py
```diff
@@ -134,6 +134,13 @@
     [58.5008586442, 28.2850942049],
 ])
 
+round_moments_expected = np.array([# sigma, e1, e2, flux, x, y
```
PaulPrice: flake8 prolly wants you to have a couple of spaces before the `#`.

tests/testHsm.py
```diff
@@ -306,6 +313,31 @@ def testHsmSourceMoments(self):
             self.assertAlmostEqual(xy, expected.getIxy(), SHAPE_DECIMALS)
             self.assertAlmostEqual(yy, expected.getIyy(), SHAPE_DECIMALS)
 
+    def testHsmRoundMoments(self):
+        for (i, imageid) in enumerate(file_indices):
+            source = self.runMeasurement("ext_shapeHSM_HsmRoundMoments", imageid, x_centroid[i], y_centroid[i],
```
PaulPrice: I _think_ this slightly exceeds the line length limit.

tests/testHsm.py
```diff
@@ -306,6 +313,31 @@ def testHsmSourceMoments(self):
...
+                                         sky_var[i])
+            x = source.get("ext_shapeHSM_HsmRoundMoments_x")
+            y = source.get("ext_shapeHSM_HsmRoundMoments_y")
+            xx = source.get("ext_shapeHSM_HsmRoundMoments_xx")
+            yy = source.get("ext_shapeHSM_HsmRoundMoments_yy")
+            xy = source.get("ext_shapeHSM_HsmRoundMoments_xy")
+            flux = source.get("ext_shapeHSM_HsmRoundMoments_Flux")
+
+            # Centroids from GalSim use the FITS lower-left corner of 1,1
```
PaulPrice: Shouldn't we correct for that in the algorithm, to present a unified set of coordinates in the catalog?

tests/testHsm.py
```diff
@@ -134,6 +134,13 @@
     [58.5008586442, 28.2850942049],
 ])
 
+round_moments_expected = np.array([# sigma, e1, e2, flux, x, y
+    [2.40270376205, 0.197810277343, -0.372329413891, 3740.22436523, 36.4032272633, 20.4847916447],
+    [1.89714717865, 0.046496052295, -0.0987404286861, 776.709594727, 20.2893584046, 25.4230368047],
+    [1.77995181084, 0.0416346564889, -0.143147706985, 534.59197998, 9.51994111869, 12.6250775205],
+    [1.46549296379, -0.0831127092242, -0.0628845766187, 348.294403076, 20.6242279632, 39.5941625731],
+    [1.64031589031, 0.0867517963052, 0.0940798297524, 793.374450684, 58.4728765002, 28.2686937854],
+])
```
PaulPrice, line 143: I think flake8 is going to want two blank lines between this and the following `def`.

PaulPrice (commented):
There are several minor stylistic issues, and I would encourage you to re-think the naming and approach.

config/enable.py
rearmstr: I guess I wasn't a fan of the original naming scheme, but I see your point.  I think ```HsmSourceMomentsRound``` is the best description.  However, I'm not a fan of the long names we have in the current catalogs, but it is probably best to be consistent.

include/lsst/meas/extensions/shapeHSM/HsmAdapter.h
rearmstr: I was waiting for a tagged GalSim version which was tagged last night.  I'll update the ticket.

tests/testHsm.py
rearmstr: No, the issue here is the difference in convention between LSST and galsim.  By default, the bounds of an image are constructed in galsim with (1,1) in the lower-left corner while LSST uses (0,0).  The precomputed galsim values we have in the test were measured using this default.  If we wanted to change this, we could change the bounds of the LSST image that is created in the test to match galsim or recompute the galsim values, but I don't think it is worth the effort.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
rearmstr: It is actually the same algorithm in galsim, just with different parameters.  I wanted this to be reflected in the LSST version.  But we want to have the results of both algorithms in the output catalogs.  So an alternative would be to create a second copy of ```HsmSrcMoments``` in the python layer and then set the appropriate parameters in a config file.  However, people could potentially call the algorithm without setting the parameters and get incorrect results.

PR #25, opened 2018-08-29
update GalSim C++ API to GalSim version 2.0.6

danielsf:


rearmstr (approved):

include/lsst/meas/extensions/shapeHSM/HsmAdapter.h
```diff
@@ -64,7 +64,7 @@ class ImageConverter
 
 private:
     PTR(afw::image::Image<PixelT>) _image;
-    boost::shared_ptr<PixelT> _owner;
+    shared_ptr<PixelT> _owner;
```
PaulPrice: `std::shared_ptr`?
And is the `#include "boost/shared_ptr.hpp"` necessary?

danielsf: I'll test it without the `#include boost/shared_ptr.hpp`

danielsf: Removing the `include` is fine, but declaring `_owner` as `std::shared_ptr<PixelT>` results in
```
:::::  [2018-08-30T19:34:48.406716Z] /local/lsst/danielsf/lsstsw3/stack/Linux64/galsim/tickets.DM-15553-g5d61ddbe65/include/galsim/Image.h:458:9: note: candidate: galsim::ImageView<T>::ImageView(const galsim::ImageView<T>&) [with T = double]
:::::  [2018-08-30T19:34:48.406723Z]          ImageView(const ImageView<T>& rhs) : BaseImage<T>(rhs) {}
:::::  [2018-08-30T19:34:48.406729Z]          ^~~~~~~~~
:::::  [2018-08-30T19:34:48.406735Z] /local/lsst/danielsf/lsstsw3/stack/Linux64/galsim/tickets.DM-15553-g5d61ddbe65/include/galsim/Image.h:458:9: note:   candidate expects 1 argument, 5 provided
:::::  [2018-08-30T19:34:48.406742Z] /local/lsst/danielsf/lsstsw3/stack/Linux64/galsim/tickets.DM-15553-g5d61ddbe65/include/galsim/Image.h:448:9: note: candidate: galsim::ImageView<T>::ImageView(T*, const std::tr1::shared_ptr<_Tp>&, int, int, const galsim::Bounds<int>&, int) [with T = double]
:::::  [2018-08-30T19:34:48.406806Z]          ImageView(T* data, const shared_ptr<T>& owner, int step, int stride,
:::::  [2018-08-30T19:34:48.406815Z]          ^~~~~~~~~
:::::  [2018-08-30T19:34:48.406821Z] /local/lsst/danielsf/lsstsw3/stack/Linux64/galsim/tickets.DM-15553-g5d61ddbe65/include/galsim/Image.h:448:9: note:   no known conversion for argument 2 from 'const std::shared_ptr<double>' to 'const std::tr1::shared_ptr<double>&'
:::::  [2018-08-30T19:34:49.506742Z] scons: *** [src/HsmMoments.os] Error 1
:::::  [2018-08-30T19:34:57.900746Z] scons: building terminated because of errors.

```
Do we want to interrogate this, or just accept that it works with `_owner` as a `shared_ptr<PixelT>`?

PaulPrice: How do you get `std::tr1::shared_ptr` in there?! Are you compiling without `-std=c++14`? What compiler?

danielsf: I am using `gcc` version 6.3.1.  I enable `devtoolset-6` to get the C++ 14 options.

danielsf: (really I just run `rebuild` in `lsstsw`)

PaulPrice: Ah, well. I don't understand it but I don't have time to run it down, so let's leave it.

TallJimbo: The `std::tr1::shared_ptr` is coming from GalSim; it has probably detected the wrong compiler and is using tr1 because it doesn't think C++11 is available.  This is definitely something we should fix now.

danielsf: So... we will know this is fixed when declaring `_owner` as a `std::shared_ptr` doesn't break?

TallJimbo: Yes (assuming my analysis of your error message is correct).

There may also be something in the GalSim build logs to indicate what compiler and standard version it is assuming.

danielsf: Is there an environment variable where `lsst_build` stores the correct compiler?  I can try just passing that into the `SCONSFLAGS` for GalSim

danielsf: The GalSim build log says
```
Using compiler: /opt/rh/devtoolset-6/root/usr/bin/g++
[2018-08-29T16:09:24.678532Z] compiler version: 6.3.1
```
That is the came compiler I get when I run `which g++`

TallJimbo: This is apparently where GalSim is making these decisions:

https://github.com/GalSim-developers/GalSim/blob/releases/2.0/include/galsim/Std.h#L51

TallJimbo: (sorry for so many posts)

The logic there basically says to use `std::shared_ptr` only if the C++ standard library is provided by Clang (libc++) rather than gcc (libstdc++).  I can think of two reasons to do that:
 - *On OS X*, gcc is ancient and hence probably doesn't provide a C++11 libstdc++.
 - New versions of libstdc++ are binary-incompatible with newer versions, which can be a real pain for people who mix old and new compilers.

The former would just be an incorrect assumption (i.e. it's just a property of OS X).  The latter would be a reasonable precaution for many GalSim users but not a problem for LSST, since we have other reasons to require all C++ code to use the newer standard library.

danielsf: It looks like I just need to define `_LIBCPP_VERSION`...?  It doesn't look like it is testing for a particular value

TallJimbo: Defining `_LIBCPP_VERSION` is probably a bit dangerous; I think only libc++ would be expected to actually do that.  We probably need @rmjarvis' opinion on this.  Now that I understand what's going on, using tr1 here for LSST wouldn't be the end of the world, but if it's a GalSim bug I bet he'll want to fix it, and it'd certainly be better not to fall back to tr1 if we can find a safe way to convince GalSim that `std::shared_ptr` is available.

danielsf: I will play with this

danielsf: Sorry.  Our messages crossed.  I thought I could get `libc++` included if I added an extra `#include` at the top of this file but, as your detailed post predicted, I could not.  I will wait for Mike to weigh-in.

PaulPrice: I believe `libc++` is a clang thing. If you're using gcc, I think you get `libstdc++`.

rmjarvis: Definitely, if the logic there is wrong, I would like to fix it.  It's possible that we've been using std::tr1::shared_ptr in places where we shouldn't have been.  I linked to a stack overflow page where I got the logic, so it's certainly possible that this logic is just bad advice and I should be doing something different there.

rmjarvis: e.g. maybe we want to add
```
#elif __cplusplus > 199711L

#include <memory>
using std::shared_ptr;
```

TallJimbo: I think the value you need for `__cplusplus` is `201103L`.

That would definitely be the most standards-conformant way to do it; I don't know how well it would work in practice for medium-old compilers that only partially support the C++11 standard, but I'd guess that `std::shared_ptr` was one of the first things in C++11 such compilers supported, so it's probably safe.

rmjarvis: I believe that `__cplusplus > 199711L` and `__cplusplus >= 201103L` should be functionally equivalent.  (Perhaps you're right that the latter would be clearer as to intent though.)

rmjarvis: I did some testing, and I'm pretty sure this works on gcc 4.8 (so presumably also anything later).  And also that it was indeed using `std::tr1::shared_ptr` previously with g++, even if `-std=c++11`.  

So I'm confident this is the right change to make.  I went ahead and tagged GalSim v2.0.7 with this change.

danielsf: Thanks, Mike.  I will incorporate that into the LSST stack and see what happens

danielsf:
The update to GalSim 2.0.7 allowed me to declare `_owner` as a `std::shared_ptr`.

The PRs in both `lsst/galsim` and here now reflect the new version.  I'm not sure who wants to be in charge of signing off on the changes, but if no one yells "stop" by noon PDT tomorrow (Friday 8/31), I will go ahead and merge (assuming that Jenkins passes ci_hsc, of course)

PR #32, opened 2021-03-25
DM-29314: Add debiased PSF moments to HSM

jmeyers314:
Debiased PSF moments are computed by rescaling flux and adding noise to
the PSF image before measuring normally.

### DIFF
--- a/include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
+++ b/include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
@@ -30,6 +30,7 @@
 #include "lsst/meas/base/CentroidUtilities.h"
 #include "lsst/meas/base/FlagHandler.h"
 #include "lsst/meas/base/InputUtilities.h"
+#include "lsst/afw/detection/Psf.h"
 
 namespace lsst { namespace meas { namespace extensions { namespace shapeHSM {
 
@@ -69,10 +70,50 @@ public:
 
 class HsmPsfMomentsControl {
 public:
-    HsmPsfMomentsControl() {}
+    LSST_CONTROL_FIELD(
+        useSourceCentroidOffset,
+        bool,
+        "Include subpixel source centroid offset when drawing PSF to be measured?"
+    );
 
-    HsmPsfMomentsControl(HsmPsfMomentsControl const & other) {}
+    HsmPsfMomentsControl() : useSourceCentroidOffset(false) {}
+
+    HsmPsfMomentsControl(HsmPsfMomentsControl const & other) :
+        useSourceCentroidOffset(other.useSourceCentroidOffset) {}
 };
+
+
+class HsmPsfMomentsDebiasedControl : public HsmPsfMomentsControl {
+public:
+    LSST_CONTROL_FIELD(
+        noiseSource,
+        std::string,
+        "Noise source.  How to choose variance of the zero-mean Gaussian noise added to image.\n"
+        "Allowed values:\n"
+        "  'meta': variance = the 'BGMEAN' metadata entry\n"
+        "  'variance': variance = the image's variance plane'\n"
+    );
+    LSST_CONTROL_FIELD(seedOffset, int, "Random seed offset");
+    LSST_CONTROL_FIELD(badMaskPlanes, std::vector<std::string>, "Mask planes used to reject bad pixels.");
+
+    HsmPsfMomentsDebiasedControl() :
+        HsmPsfMomentsControl(),
+        noiseSource("variance"),
+        seedOffset(0)
+    {
+        useSourceCentroidOffset=true;
+        badMaskPlanes.push_back("BAD");
+        badMaskPlanes.push_back("SAT");
+    }
+
+    HsmPsfMomentsDebiasedControl(HsmPsfMomentsDebiasedControl const & other) :
+        HsmPsfMomentsControl(other),
+        noiseSource(other.noiseSource),
+        seedOffset(other.seedOffset),
+        badMaskPlanes(other.badMaskPlanes)
+    {}
+};
+
 
 /// Base class to measure HSM adaptive moments
 ///
@@ -114,7 +155,8 @@ protected:
         afw::image::MaskPixel const badPixelMask, // Bitmask for bad pixels
         float const width,            // PSF width estimate, for starting moments
         bool roundMoments=false, // Use round weight function
-        bool addFlux=false // add Flux to output
+        bool addFlux=false, // add Flux to output
+        bool subtractCenter=false // subtract starting center from x/y outputs?
     ) const;
 
 public:
@@ -158,12 +200,13 @@ private:
     Control _ctrl;
 };
 
+
 /// Class to measure HSM adaptive moments of PSF
 class HsmPsfMomentsAlgorithm : public HsmMomentsAlgorithm {
 public:
     /// A typedef to the Control object for this algorithm, defined above.
     ///  The control object contains the configuration parameters for this algorithm.
-    typedef HsmPsfMomentsControl Control;
+    typedef std::shared_ptr<const HsmPsfMomentsControl> Control;
 
     /// @brief Initialize with standard field names and customized documentation.
     HsmPsfMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
@@ -174,8 +217,59 @@ public:
         afw::image::Exposure<float> const & exposure
     ) const;
 
-private:
+protected:
+    virtual std::shared_ptr<afw::detection::Psf::Image> getPsfImage(
+        geom::Point2D center,
+        afw::table::SourceRecord & source,
+        afw::image::Exposure<float> const & exposure
+    ) const;
+
+    virtual std::shared_ptr<afw::image::Mask<afw::image::MaskPixel>> getPsfMask(
+        geom::Point2D center,
+        afw::table::SourceRecord & source,
+        afw::image::Exposure<float> const & exposure
+    ) const;
+
+    virtual afw::image::MaskPixel const getBadPixelMask(
+        afw::image::Exposure<float> const & exposure
+    ) const { return 0; }
+
     Control _ctrl;
+};
+
+
+/// Measure PSF moments while making image signal-to-noise look similar to underlying star,
+/// thereby removing any signal-to-noise related biases.
+class HsmPsfMomentsDebiasedAlgorithm : public HsmPsfMomentsAlgorithm {
+public:
+    static base::FlagDefinition const EDGE;
+
+    typedef std::shared_ptr<const HsmPsfMomentsDebiasedControl> Control;
+
+    HsmPsfMomentsDebiasedAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
+        HsmPsfMomentsAlgorithm(ctrl, name, schema) {
+            auto thisCtrl = std::static_pointer_cast<Control::element_type>(_ctrl);
+            if ((thisCtrl->noiseSource != "meta") && (thisCtrl->noiseSource != "variance")) {
+                throw LSST_EXCEPT(base::MeasurementError, "invalid noiseSorce", FAILURE.number);
+            }
+        }
+
+private:
+    std::shared_ptr<afw::detection::Psf::Image> getPsfImage(
+        geom::Point2D center,
+        afw::table::SourceRecord & source,
+        afw::image::Exposure<float> const & exposure
+    ) const override;
+
+    std::shared_ptr<afw::image::Mask<afw::image::MaskPixel>> getPsfMask(
+        geom::Point2D center,
+        afw::table::SourceRecord & source,
+        afw::image::Exposure<float> const & exposure
+    ) const override;
+
+    afw::image::MaskPixel const getBadPixelMask(
+        afw::image::Exposure<float> const & exposure
+    ) const override;
 };
 }}}} // namespace lsst::meas::extensions::shapeHSM
 
--- a/include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
+++ b/include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
@@ -107,8 +107,8 @@ protected:
     template<typename PixelT>
     void calculate(
         afw::table::SourceRecord& source, // Source for recording moments
-        PTR(afw::image::Image<PixelT>) const& afwImage, // Image on which to measure moments
-        PTR(afw::image::Mask<afw::image::MaskPixel>) const& afwMask, // Mask for image
+        std::shared_ptr<afw::image::Image<PixelT>> const& afwImage, // Image on which to measure moments
+        std::shared_ptr<afw::image::Mask<afw::image::MaskPixel>> const& afwMask, // Mask for image
         geom::Box2I const& bbox,     // Bounding box
         geom::Point2D const& center, // Starting center for measuring moments
         afw::image::MaskPixel const badPixelMask, // Bitmask for bad pixels
--- a/src/HsmMoments.cc
+++ b/src/HsmMoments.cc
@@ -27,6 +27,7 @@
 #include "lsst/afw/math/Statistics.h"
 #include "lsst/afw/table/Source.h"
 #include "lsst/afw/geom/ellipses.h"
+#include "lsst/afw/math/Random.h"
 
 #include "galsim/Image.h"
 #include "galsim/hsm/PSFCorr.h"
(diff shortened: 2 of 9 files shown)


include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -69,9 +69,24 @@ class HsmSourceMomentsRoundControl : public HsmSourceMomentsControl {
 
 class HsmPsfMomentsControl {
 public:
-    HsmPsfMomentsControl() {}
+    LSST_CONTROL_FIELD(debias, bool, "Debias PSF moments?");
+    LSST_CONTROL_FIELD(noiseSource, std::string, "Noise source");
```
TallJimbo: This entry definitely needs more documentation, and having seen how it's used, maybe call it `noiseVarianceSource`, so it's harder to mistake for e.g. a RNG seed?

src/HsmMoments.cc
```diff
@@ -145,15 +146,56 @@ void HsmPsfMomentsAlgorithm::measure(
     geom::Point2D center = _centroidExtractor(source, _flagHandler);
 
     typedef afw::image::Mask<afw::image::MaskPixel> Mask;
-    PTR(afw::detection::Psf::Image) image = exposure.getPsf()->computeKernelImage(center);
+    PTR(afw::detection::Psf::Image) psfImage;
```
TallJimbo: The `PTR` macro is deprecated; use
```
std::shared_ptr<afw::detector::Psf::Image> psfImage;
```

src/HsmMoments.cc
```diff
@@ -145,15 +146,56 @@ void HsmPsfMomentsAlgorithm::measure(
...
+
+    if (_ctrl.debias) {
+        typedef afw::detection::Psf::Pixel PsfPixel;
```
TallJimbo: Preferred style for this kind of typedef is:
```
using afw::detection::Psf::Pixel;
```

src/HsmMoments.cc
```diff
@@ -145,15 +146,56 @@ void HsmPsfMomentsAlgorithm::measure(
...
+
+        // draw Psf with same offset as star
+        psfImage = exposure.getPsf()->computeImage(center);
```
TallJimbo: This (drawing PSF with the same offset as the star) is an interesting choice.  It might be the right one, but I think that's an experimental question (if they're not equivalent, it says something about our interpolation that would matter in other places).  And I think that you probably need to go tell the moments algorithm later to start from that fractional centroid other than `(0, 0)`.  While I think the moments algorithm is capable of adjusting its center a bit, I'm worried that it may not be able to compensate enough.

So, for now, can we put this behind a config option, in addition to adjusting the center below?  I don't care much what the default config is, but I think we'll want to try both offset choices without recompiling.

Also, if it turns out the moments algorithm updates its center even less than I expect _and_ using the star position/`computeImage` matters, we may want to start computing a PSF-model centroid variant in `SdssCentroid`, too, so we can use a centroid here that includes measurement errors.  But I'm hoping that won't be necessary.

src/HsmMoments.cc
```diff
@@ -145,15 +146,56 @@ void HsmPsfMomentsAlgorithm::measure(
...
+            _flagHandler.setValue(source, FAILURE.number, true);
+        }
+        (*psfImage) *= flux;
+
+        // Add Gaussian noise to image from variance plane
+        afw::image::Image<PsfPixel> noise(fitBBox);
+        afw::math::Random rand(afw::math::Random::MT19937, source.getId());
+        afw::math::randomGaussianImage<afw::image::Image<PsfPixel>>(&noise, rand);
+        if (_ctrl.noiseSource == "meta") {
+            double bgmean = exposure.getMetadata()->getAsDouble("BGMEAN");  // Should I catch NotFoundError?
```
TallJimbo: > Should I catch NotFoundError?

Yes, and probably re-raise as `FatalAlgorithmError`, because if it's not set for one image in a processing run, it's probably not going to be set for many of them, and that's a configuration error the operator probably wants to fix.

src/HsmMoments.cc
```diff
@@ -145,15 +146,56 @@ void HsmPsfMomentsAlgorithm::measure(
...
+        geom::Box2I fitBBox = psfImage->getBBox();
+        fitBBox.clip(exposure.getBBox());
+        if (fitBBox != psfImage->getBBox()) {
+            _flagHandler.setValue(source, FAILURE.number, true);
+        }
+        (*psfImage) *= flux;
+
+        // Add Gaussian noise to image from variance plane
+        afw::image::Image<PsfPixel> noise(fitBBox);
+        afw::math::Random rand(afw::math::Random::MT19937, source.getId());
```
TallJimbo: Our standard practice is to also have a config field that provides a constant offset to any deterministic noise seed like `source.getId`, so we can have two otherwise-identical runs with different random numbers.

src/HsmMoments.cc
```diff
@@ -145,15 +146,56 @@ void HsmPsfMomentsAlgorithm::measure(
...
+            afw::image::Image<float> var(
+                *exposure.getMaskedImage().getVariance(),
+                fitBBox,
+                afw::image::PARENT,
+                true
+            );
+            var.sqrt();
+            noise *= var;
+        } else {
+            throw LSST_EXCEPT(base::MeasurementError, "invalid noiseSorce", FAILURE.number);
```
TallJimbo: This check needs to move to the algorithm constructor so it can fail before any processing starts (we can't always do that, but whenever we can, it's much better).

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -69,9 +69,24 @@ class HsmSourceMomentsRoundControl : public HsmSourceMomentsControl {
 
 class HsmPsfMomentsControl {
 public:
-    HsmPsfMomentsControl() {}
+    LSST_CONTROL_FIELD(debias, bool, "Debias PSF moments?");
```
TallJimbo: I'm not a fan of the fact that this makes it possible to turn the debiasing off the plugin we've named "debiased".

I think it would be a lot cleaner to split the `HsmPsfMomentsAlgorithm` class into a base + 2 derived triplet, where the new base class (best to keep the name of the not-debiased concrete algorithm the same) implements `measure` by calling a new `getPsfImage` virtual method that each derived class would implement.

TallJimbo (changes requested):

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
jmeyers314: Will do.  FWIW, the name matches a similar config field in `NoiseReplacer`: https://github.com/lsst/meas_base/blob/master/python/lsst/meas/base/noiseReplacer.py#L35.  I suspect that once I add a field for a seed offset, it'll be clearer that this is not a seed.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
jmeyers314: Sounds good.  I haven't looked in detail yet, but I think the same comment applies to `HsmSourceMomentsRound`?  I.e., I could manipulate its control to set round=false?  If so I'll refactor that one too.

TallJimbo: Yes, please do.  I had forgotten all about that one.

src/HsmMoments.cc
jmeyers314: > So, for now, can we put this behind a config option, in addition to adjusting the center below?

I'm happy to put these behind a config option, but do you have a suggestion for how to interpolate the variance plane in the case where `computeKernelImage` is used?  Nearest pixel is easy enough, but maybe there's code somewhere else in the stack that could do something better?

TallJimbo: > Nearest pixel is easy enough, but maybe there's code somewhere else in the stack that could do something better?

Good point; I hadn't thought about what to do about the variance plane when proposing this.  There are better interpolants, but I'm not super familiar with it, and it'll bring some complications with it (e.g. edge effects).  Let's just do nearest pixel for now and document that as a limitation of that config choice.

jmeyers314: > This (drawing PSF with the same offset as the star) is an interesting choice. It might be the right one, but I think that's an experimental question (if they're not equivalent, it says something about our interpolation that would matter in other places).

Looking into this a bit more.  I decided it would be worth trying out drawing an offset PSF even in the no-additional-noise-added case, and have found that, indeed, the current unit tests fail in this case.  Specifically, getting the HSM _centroid_ of a GaussianPsf that's drawn off-center still works fine, but the HSM _second moments_ are affected.  For a half-pixel shift with expected Ixx=4.0 pix^2, HSM yields 3.988, so I'd need to reduce the test tolerance from 3 decimals to 1.  (For a centered PSF image, HSM actually gives the expected result to 5 decimals).

I suspect this says more about HSM though than about PSF interpolation.  But it definitely seems like a source of bias when comparing PSFs and stars.

jmeyers314: > I suspect this says more about HSM though than about PSF interpolation. 

I take it back, as when I try to reproduce this using just GalSim, HSM seems perfectly able to deal with pixel offsets.  Is `GaussianPsf.computeImage()` implemented by direct computation or by Lanczosing the `computeKernelImage()` result?  If it's the latter, then yeah, that could be a problem.  The next question is then what do `Psfex` and `Piff` do?  Piff at least I know draws the offset image directly (though there is certainly interpolation under the hood, it's fundamentally part of the model, not something added on top).

```python
gaussian = galsim.Gaussian(sigma=2.0)

img = gaussian.drawImage(nx=35, ny=35, scale=1.0, method='no_pixel')
mom = galsim.hsm.FindAdaptiveMom(img)
print(mom.moments_centroid)
print(mom.moments_sigma)
print(mom.observed_shape)

# draw offset directly
img2 = gaussian.drawImage(nx=35, ny=35, scale=1.0, method='no_pixel', offset=(0.5, 0.0))
mom2 = galsim.hsm.FindAdaptiveMom(img2)
print()
print(mom2.moments_centroid)
print(mom2.moments_sigma)
print(mom2.observed_shape)

# Use Lanczos magic to draw with offset
ii = galsim.InterpolatedImage(img, scale=1.0, normalization='sb', x_interpolant='Lanczos3')
img3 = ii.drawImage(nx=35, ny=35, scale=1.0, method='no_pixel', offset=(0.5, 0.0))
mom3 = galsim.hsm.FindAdaptiveMom(img3)
print()
print(mom3.moments_centroid)
print(mom3.moments_sigma)
print(mom3.observed_shape)
```

yields

```
galsim.PositionD(18.0,18.0)
2.0
galsim.Shear(g1=5.551115123125783e-17,g2=9.721081979824222e-18)

galsim.PositionD(18.5,18.0)
2.0
galsim.Shear(g1=-6.354167192412774e-10,g2=1.1135718818786897e-17)

galsim.PositionD(18.5,18.0)
1.997077465057373
galsim.Shear(g1=-0.0014623346344376938,g2=1.5930238952803163e-17)
```

TallJimbo: I'm pretty sure `CoaddPsf` is just going to use a Lanczos kernel to interpolate from `computeKernelImage` to `computeImage`, and yes, I've been worried about that as a potential problem for a while - though I _think_ someone tried the experiment of upping the Lanczos order there at some point and found that didn't explain the difference between our single-visit rho statistics and coadd rho statistics.  We should definitely return to that question after we have a consistent sample of reserved stars.

PSFEx would do its own interpolation, and I think that entails a Lanczos interpolation of the upsampled model image, but I'm not certain.

jmeyers314: Sounds good.  

I think the immediate question is: should I fix `afwDetection::GaussianPsf.computeImage()` now to not use Lanczos so that new HSM unit tests pass with tight thresholds, or should I just use really loose thresholds on those new tests?  I opted for loose thresholds for now, but I don't think it'd actually be that much work to fixup GaussianPsf.

TallJimbo: I think the loose thresholds are fine here, and I suspect modifying `GaussianPsf` might would perturb a bunch of existing tests in a way that would be just a waste of time to deal with.

jmeyers314: I found a third option here, which was to implement a pure-python Gaussian Psf that handles `computeImage` exactly.  Thanks to trampolines this is only ~40 lines of pure python now :)

jmeyers314:
Thanks for the initial review @TallJimbo .  I think I’ve addressed all your comments now.  I did change some other things along the way though, so you probably should look at the whole PR again.  Here are the things I updated that weren’t part of your initial review:

- Added a `doCenter=True` option to `HsmPsfMomentsControl` so it can be used both for the original and debiased PSF algorithms.
- Added a `badMaskPlanes` control option to the debiased psf algorithm.  The motivation here is to enable Psf measurements and star measurements to be as close to the same as possible.
- Added an EDGE flag to the debiased psf algorithm.  Set if the PSF overlaps the image edge.  The algorithm still runs unless it needs variance inputs from off the edge, in which case it raises a base::MeasurementError.
- I went a slightly different direction for Psf / PsfDebiased hierarchy.  Instead of a triplet, the debiased algorithm just inherits from the regular algorithm.  The debiased algorithm can use some of the regular algorithm’s methods this way.  Let me know if there’s a downside to this design though.
- To make the above work, I wrapped the algorithm’s control object in a shared_ptr, so there are some static_pointer_casts in the debiased methods.
- I didn’t refactor the roundMoments stuff yet (I wanted to get your reaction to my Psf design first).  One possible quick-fix that occurred to me was to just delete the `LSST_DECLARE_CONTROL_FIELD` for the roundMoments in python/…/hsmMomentsControl.cc.  I think this would make that field inaccessible from python at least.  I don’t know if it’d break something else like config <-> control conversions though.
- Finally, I left sloppy fixit-style commits here for the moment.  I'll rebase before merge though :)

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -69,11 +70,47 @@ class HsmSourceMomentsRoundControl : public HsmSourceMomentsControl {
 
 class HsmPsfMomentsControl {
 public:
-    HsmPsfMomentsControl() {}
+    LSST_CONTROL_FIELD(doCenter, bool, "Center PSF image before measuring?");
```
TallJimbo: It's not obvious to me which of the two options we discussed corresponds to true and false here.  How about `useSourceCentroidOffset` (true for `computeImage`, false for `computeKernelImage`)?

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -158,25 +196,81 @@ class HsmSourceMomentsAlgorithm : public HsmMomentsAlgorithm {
...
     /// A typedef to the Control object for this algorithm, defined above.
     ///  The control object contains the configuration parameters for this algorithm.
-    typedef HsmPsfMomentsControl Control;
+    typedef std::shared_ptr<const HsmPsfMomentsControl> Control;
 
     /// @brief Initialize with standard field names and customized documentation.
     HsmPsfMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
         HsmMomentsAlgorithm(name, schema, "Psf adaptive moments algorithm from HSM"), _ctrl(ctrl) {}
 
+    virtual ~HsmPsfMomentsAlgorithm() {}
```
TallJimbo: Should be unnecessary; the ultimate base class destructor is already `virtual`.  If we did define it here I think we'd want to say
```
virtual ~HsmPsfMomentsAlgorithm() noexcept = default;
```

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -158,25 +196,81 @@ class HsmSourceMomentsAlgorithm : public HsmMomentsAlgorithm {
...
 };
 
+
 /// Class to measure HSM adaptive moments of PSF
 class HsmPsfMomentsAlgorithm : public HsmMomentsAlgorithm {
 public:
     /// A typedef to the Control object for this algorithm, defined above.
     ///  The control object contains the configuration parameters for this algorithm.
-    typedef HsmPsfMomentsControl Control;
+    typedef std::shared_ptr<const HsmPsfMomentsControl> Control;
```
TallJimbo, line 209: Was this necessary to appease pybind11?  It's not wrong, I guess, but passing control objects by `shared_ptr` is just not something I would have considered.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -158,25 +196,81 @@ class HsmSourceMomentsAlgorithm : public HsmMomentsAlgorithm {
...
+        }
+
+    virtual ~HsmPsfMomentsDebiasedAlgorithm() {}
+
+private:
+    virtual std::shared_ptr<afw::detection::Psf::Image> getPsfImage(
+        geom::Point2D center,
+        afw::table::SourceRecord & source,
+        afw::image::Exposure<float> const & exposure
+    ) const override;
```
TallJimbo, line 262: Please drop the `virtual` from this and the methods below it: https://developer.lsst.io/cpp/style.html#the-override-specifier-should-be-used-whenever-the-intention-is-to-override-a-function-from-a-base-class

TallJimbo (approved):
Looks good!  Only minor comments this time around.  Sorry it took me so long to get back to it.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
jmeyers314: This wasn't for the argument passing (or anything pybind11), but to enable one class hierarchy to hold onto instances of a different class hierarchy.  I.e., I needed a pointer instead of a concrete object so that `HsmPsfMomentsAlgorithm` could hold a `HsmPsfMomentsControl` and `HsmPsfMomentsDebiasedAlgorithm` could hold a `HsmPsfMomentsDebiasedControl`, and the debiased methods that call the parent methods still work.  If there's a different pattern you recommend, definitely let me know!

TallJimbo: Makes sense, and I don't have a better idea :+1:.

PR #33, opened 2021-04-23
DM-29863: ShapeHSM fails when psf bbox size is position dependent

jmeyers314:


tests/test_hsm.py
```diff
@@ -384,8 +390,11 @@ def testHsmPsfMoments(self):
                     (22.81, 34.01),
                     (22.81, 33.99),
                     (1.2, 1.3),  # psfImage extends outside exposure; that's okay
+                    (-100.0, -100.0),
+                    (-100.5, -100.0),
+                    (-100.5, -100.5),
```
erykoff, line 395: To be clear, this is testing that the negative positions work fine after you've put in the bbox center fix?

erykoff (approved):
Looks good.

tests/test_hsm.py
jmeyers314: Yep.  I think it's probably overkill, but certainly does no harm.

PR #35, opened 2021-06-03
DM-30426: Use psf->computeImageBBox()

jmeyers314:
In particular, avoid inferring `psf->computeImage()->getBBox()`
from `psf->computeBBox()`, since this has proven unreliable and
different Psf subclasses may even perform the conversion
differently.

### DIFF
--- a/src/HsmMoments.cc
+++ b/src/HsmMoments.cc
@@ -154,16 +154,24 @@ void HsmPsfMomentsAlgorithm::measure(
 ) const {
     geom::Point2D center = _centroidExtractor(source, _flagHandler);
 
+    auto psf = exposure.getPsf();
+    if (!psf) {
+        throw LSST_EXCEPT(base::MeasurementError, "Exposure lacks PSF", FAILURE.number);
+    }
+
+    // Goal below is to get an image+mask of the psf with a bbox like
+    // computeImageBBox(), even when useSourceCentroidOffset is False.
     auto psfImage = getPsfImage(center, source, exposure);
     auto psfMask = getPsfMask(center, source, exposure);
     auto badPixelMask = getBadPixelMask(exposure);
 
-    double const sigmaGuess = exposure.getPsf()->computeShape(center).getTraceRadius();
-    const geom::Point2D centroidGuess = (
-        _ctrl->useSourceCentroidOffset ?
-        center :
-        geom::Point2D(std::floor(center.getX()+0.5), std::floor(center.getY()+0.5))
-    );
+    double const sigmaGuess = psf->computeShape(center).getTraceRadius();
+
+    geom::Point2D centroidGuess = center;
+    if (!_ctrl->useSourceCentroidOffset) {
+        auto bbox = psf->computeImageBBox(center);
+        centroidGuess = geom::Point2D(bbox.getMin() + bbox.getDimensions()/2);
+    }
 
     bool roundMoments = false;
     bool addFlux = false;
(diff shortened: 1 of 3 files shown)


src/HsmMoments.cc
```diff
@@ -158,12 +158,14 @@ void HsmPsfMomentsAlgorithm::measure(
...
-    const geom::Point2D centroidGuess = (
-        _ctrl->useSourceCentroidOffset ?
-        center :
-        geom::Point2D(std::floor(center.getX()+0.5), std::floor(center.getY()+0.5))
-    );
+    auto psf = exposure.getPsf();
+    double const sigmaGuess = psf->computeShape(center).getTraceRadius();
+
+    geom::Point2D centroidGuess = center;
+    if (!_ctrl->useSourceCentroidOffset) {
```
PaulPrice, line 171: You need to document what `useSourceCentroidOffset` is for somewhere, at the very least in a code comment or a git commit message. But I don't see why you would ever want `useSourceCentroidOffset=False`. The fact that we used `computeKernelImage` instead of `computeImage` in the past was just for convenience, and if that has the potential to affect the results then I don't think there's any reason to stick with it. I think you should simply get rid of `useSourceCentroidOffset`, which greatly simplifies things.

src/HsmMoments.cc
```diff
@@ -158,12 +158,14 @@ void HsmPsfMomentsAlgorithm::measure(
     auto psfMask = getPsfMask(center, source, exposure);
     auto badPixelMask = getBadPixelMask(exposure);
 
-    double const sigmaGuess = exposure.getPsf()->computeShape(center).getTraceRadius();
```
PaulPrice, line 161: Why are the changes to `HsmPsfMomentsAlgorithm::measure` related to declared goal of the commit, i.e., "Use psf->computeImageBBox()"?

src/HsmMoments.cc
```diff
@@ -179,14 +181,13 @@ std::shared_ptr<afw::detection::Psf::Image> HsmPsfMomentsAlgorithm::getPsfImage(
     afw::table::SourceRecord & source,
     afw::image::Exposure<float> const & exposure
 ) const {
+    auto psf = exposure.getPsf();
```
PaulPrice, line 190: Probably should be a check for a `NULL` here, and other similar places, especially the `measure` method.

src/HsmMoments.cc
```diff
@@ -197,12 +198,7 @@ std::shared_ptr<afw::image::Mask<afw::image::MaskPixel>> HsmPsfMomentsAlgorithm:
     afw::image::Exposure<float> const & exposure
 ) const {
     auto psf = exposure.getPsf();
-    auto bbox = psf->computeBBox(center);
-
-    // Add floor(center+0.5) to bbox to make it look like psf->computeImage()->getBBox()
-    auto shift = geom::Extent2I(std::floor(center.getX()+0.5), std::floor(center.getY()+0.5));
-    bbox.shift(shift);
-
+    auto bbox = psf->computeImageBBox(center);
```
PaulPrice, line 207: Don't you need to use `computeKernelBBox` if `useSourceCentroidOffset=False`?

tests/test_hsm.py
```diff
@@ -382,57 +390,63 @@ class PsfMomentsTestCase(unittest.TestCase):
     def testHsmPsfMoments(self):
```
PaulPrice: I recommend use of the `lsst.utils.tests.methodParameters` function decorator.

PaulPrice (approved):
I recommend stripping out the `useSourceCentroidOffset` knob, which will simplify the code.

src/HsmMoments.cc
jmeyers314: The documentation is [here](https://github.com/lsst/meas_extensions_shapeHSM/blob/tickets/DM-30426/include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h#L76).

src/HsmMoments.cc
jmeyers314: The goal for both `getPsfImage` and `getPsfMask` is to always return something with a bbox equal to `psf->computeImageBBox(center)`, even when `useSourceCentroidOffset` is False.  (That's what the futzing around with `setXY0` is for in `getPsfImage`).  I'll add a clarifying comment to `measure`.

src/HsmMoments.cc
jmeyers314: I added a check near the beginning of `measure`.  I think that's sufficient since the other similar instances are all beneath that.

src/HsmMoments.cc
jmeyers314: I needed to change the way `centroidGuess` was being calculated to reflect the change in how the bbox was calculated.

PR #38, opened 2022-06-16
DM-35007: Make schema documentation of HSM moments useful

arunkannawadi:


include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -183,8 +183,9 @@ class HsmSourceMomentsAlgorithm : public HsmMomentsAlgorithm {
     typedef HsmSourceMomentsControl Control;
 
     /// @brief Initialize with standard field names and customized documentation.
-    HsmSourceMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
-        HsmMomentsAlgorithm(name, schema, "Source adaptive moments algorithm from HSM"), _ctrl(ctrl)
+    HsmSourceMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema,
+                              char const* doc = "Source adaptive moments algorithm from HSM") :  ///< Docstring is exposed for HsmSourceMomentsRoundAlgorithm
```
parejkoj: "from HSM" or "from shapeHSM"? Are we referring to the `HSM` paper here, or to this `shapeHSM` package?

Seeing how these come out in the schema, I wonder if "Source adaptive moments via the HSM shape algorithm" (still not sure whether to use `shapeHSM` or `HSM shape` or what) might be better? A user reading the catalog schema is asking "what is this and where did it come from", and "X algorithm from HSM" is answering a different question.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -201,6 +202,15 @@ class HsmSourceMomentsAlgorithm : public HsmMomentsAlgorithm {
 };
 
 
+class HsmSourceMomentsRoundAlgorithm : public HsmSourceMomentsAlgorithm {
+public:
+    /// @brief Initialize with standard field names and customized documentation.
+    HsmSourceMomentsRoundAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema,
+                                   char const* doc = "Source adaptive moments algorithm from HSM, with circular weight function") :
```
parejkoj: As above, "Source adaptive moments via the HSM shape algorithm..." or something like that.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -209,9 +219,10 @@ class HsmPsfMomentsAlgorithm : public HsmMomentsAlgorithm {
     typedef std::shared_ptr<const HsmPsfMomentsControl> Control;
 
     /// @brief Initialize with standard field names and customized documentation.
-    HsmPsfMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
-        HsmMomentsAlgorithm(name, schema, "Psf adaptive moments algorithm from HSM"), _ctrl(ctrl) {}
-
+    HsmPsfMomentsAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema,
+                           char const* doc = "PSF adaptive moments algorithm from HSM"  ///< Docstring is exposed for debiased PSF algorithm
```
parejkoj: As above, "PSF adaptive moments via the HSM shape algorithm"

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
```diff
@@ -246,8 +257,9 @@ class HsmPsfMomentsDebiasedAlgorithm : public HsmPsfMomentsAlgorithm {
 
     typedef std::shared_ptr<const HsmPsfMomentsDebiasedControl> Control;
 
-    HsmPsfMomentsDebiasedAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema) :
-        HsmPsfMomentsAlgorithm(ctrl, name, schema) {
+    HsmPsfMomentsDebiasedAlgorithm(Control const & ctrl, std::string const & name, afw::table::Schema & schema,
+                                   char const* doc = "Debiased PSF adaptive moments algorithm from HSM") :
```
parejkoj: As above, "Debiased PSF adaptive moments via the HSM shape algorithm", or similar.

python/lsst/meas/extensions/shapeHSM/__init__.py
```diff
@@ -39,7 +39,7 @@
                     Control=HsmShapeRegaussControl, executionOrder=BasePlugin.SHAPE_ORDER)
 wrapSimpleAlgorithm(HsmSourceMomentsAlgorithm, name="ext_shapeHSM_HsmSourceMoments",
                     Control=HsmSourceMomentsControl, executionOrder=BasePlugin.SHAPE_ORDER)
-wrapSimpleAlgorithm(HsmSourceMomentsAlgorithm, name="ext_shapeHSM_HsmSourceMomentsRound",
+wrapSimpleAlgorithm(HsmSourceMomentsRoundAlgorithm, name="ext_shapeHSM_HsmSourceMomentsRound",
```
parejkoj, line 42: Oh, that's definitely a bug! Good catch.

parejkoj (changes requested):
Thanks for improving these docs; I have some wording suggestions, based on how these look in the schema you posted on Jira. I'm not wedded to that new wording, but I think we should look at it from the perspective of someone reading the schema.

I do wish we had some discussion about what exactly HSM *is*. The readme is entirely unhelpful, and we don't have any package docs. I believe the term comes from the name of one or more papers: could you please add a link to the relevant papers to the readme as part of this? That's the minimum: I think some package docs are probably out of scope for this ticket, but please file a ticket linked from this one saying that we need them.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
plazas: Forgive me for intruding, but I thought HSM came from "Hirata, Seljak Moments": https://ui.adsabs.harvard.edu/abs/2003MNRAS.343..459H/abstract   (adaptive moments). I've heard "Hirata, Seljak, Mandelbaum", but I'm not sure that's the case. I'll ask Rachel Mandelbaum.  (This is in response to the comment above "I do wish we had some discussion about what exactly HSM is. ")

arunkannawadi:
Yup, you're correct Andres. That's the origin of HSM.

plazas:
So, it turns out that I was wrong, and it is indeed "Hirata, Seljak, and Mandelbaum" :)   I'm glad I asked Rachel. She said that Hirata & Seljak developed the formalism and initial code, but then she modified it when applying it in practice to the Mandelbaum et al (2005) SDSS shape catalog paper.  So it’s Hirata, Seljak, and Mandelbaum because they initiated the concept and code and she made updates to get it working in practice on the data.

arunkannawadi:
Oh, I misread the exact wording of your original comment. I did mean HSM
stands for Hirata, Seljak & Mandelbaum but good that you double checked
that

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
arunkannawadi: I don't think `shapeHSM` is going to be particularly useful either, since that's literally the string in the column name. HSM, in practice, refers to a very specific implementation of an algorithm that's not fundamentally different from SDSS shapes. I don't have a better alternative to what you suggested John, so I will go with it. I'll add the reference in the README and create a ticket to put in some module level documentation explaining it in a bit more detail.

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
arunkannawadi: I can see how this can be confusing as well, if it is interpreted to mean some PSF was adapted to some standard one to measure the moments. I will change this to "Adaptive moments of the PSF via the HSM shape algorithm".

include/lsst/meas/extensions/shapeHSM/HsmMomentsControl.h
arunkannawadi: and this to "Debiased adaptive moments of the PSF via the HSM shape algorithm".

python/lsst/meas/extensions/shapeHSM/__init__.py
arunkannawadi: FYI: It wasn't a bug, since whether round moments are calculated or not, was entiredly determined with the `Control` class. `HsmSourceMomentsRoundAlgorithm` is defined in this ticket, and is subclassed solely to have a different docstring.

parejkoj (approved):
Thank you, that info in the readme is a good start to describing what this does.

PR #39, opened 2022-09-14
Enable use of GalSim 2.4

jmeyers314:


PaulPrice (approved):
Please add more details in the commit message (what changed, why is this the best way to fix it?), and add a space either side of the `+`.

PR #42, opened 2023-06-04
DM-39316: Combine meas_extensions_shapeHSM pybind11 wrappers into single shared library

mwittgen:


mwittgen:
Made obsolete by #47

PR #43, opened 2023-10-25
DM-30947: Reimplement C++ HSM moment measurement algorithms in Python

enourbakhsh:


python/lsst/hsm_moments.py
```diff
@@ -0,0 +1,111 @@
...
+# import MeasurementError
+
+class HsmMomentsConfig(SingleFramePluginConfig):
+    # roundMoments
+    # addFlux
+    roundMoments = pexConfig.Field[bool](doc="Use round weight function?", default=False)
+    addFlux = pexConfig.Field[bool](doc="Store measured flux?", default=False)
+
+
+class HsmSourceMomentsConfig(HsmMomentsConfig):
```
arunkannawadi: You need to add the following decorator to the plugin
```python
@lsst.meas.base.register("ext_shapeHSM_HsmSourceMoments")
```

and remove these lines: https://github.com/lsst/meas_extensions_shapeHSM/blob/main/python/lsst/meas/extensions/shapeHSM/__init__.py#L40-L41

Similarly, you should add decorators to register the various plugins and keep the same names as used in the `__init__`. The removal of lines as you add decorators is to avoid any naming clashes.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+@measBase.register("ext_shapeHSM_HsmSourceMoments")
+class HsmSourceMomentsPlugin(HsmMomentsPlugin):
+    ConfigClass = HsmSourceMomentsConfig
+
+    def __init__(self, config, name, schema, metadata, logName=None) -> None:
+        super().__init__(config, name, schema, logName)
+        # some flag thingy
+        # work in conjunction with fail()
+        self.centroidResultKey = afwTable.Point2DKey.addFields(
+            schema, "centroid", "The centroid measured by HSM", "pixel"
```
arunkannawadi:
```suggestion
            schema, name, "The centroid measured by HSM", "pixel"
```

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            shape = galsim.hsm.FindAdaptiveMom(
+                image,
+                weight=None,
+                badpix=badPixelMask,
+                guess_sig=2.5 * psfSigma,
+                precision=1.0e-6,
+                guess_centroid=None,
+                strict=True,
+                round_moments=False,
```
arunkannawadi:
```suggestion
                round_moments=self.config.roundMoments,
```

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+        # now that we have shape, we can get the values below:
+        determinantRadius = shape.moments_sigma
+        centroidResult = shape.moments_centroid
+
+        # subtract center should be implemented as well
+        if self.config.subtractCenter:
+            centroidResult.x -= center.getX()
+            centroidResult.y -= center.getY()
+
+        record.set(self.centroidResultKey, centroidResult)
```
arunkannawadi: `centroidResult` would be an instance of a `galsim` class, and should ideally be a `lsst.geom.Point2D`. I'm surprised this didn't result in an error.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+        # make sure bbox is centered properly
+
+        try:
+            shape = galsim.hsm.FindAdaptiveMom(
+                image,
+                weight=None,
+                badpix=badPixelMask,
+                guess_sig=2.5 * psfSigma,
+                precision=1.0e-6,
+                guess_centroid=None,
```
arunkannawadi:
```suggestion
                guess_centroid=center,
```

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+        self.shapeKey = afwTable.QuadrupoleKey.addFields(schema, name, doc="HSM source moments.")
+        if config.addFlux:  # search the correct attribute name in the codebase
+            self.fluxKey = measBase.FluxResultKey.addFields(schema, "flux", "The flux measured by HSM", "dn")
+
+    def measure(self, record, exposure):
+        """... in place"""
+        # source in c++ is record in py. exposure the same
+
+        center = record.getCentroid()
```
arunkannawadi: We should switch to getting the centroid safely. See:
https://github.com/lsst/meas_extensions_convolved/blob/42fff21dfb88e9ac1772aa792458ec200baa4b58/python/lsst/meas/extensions/convolved/convolved.py#L278
and
https://github.com/lsst/meas_extensions_convolved/blob/42fff21dfb88e9ac1772aa792458ec200baa4b58/python/lsst/meas/extensions/convolved/convolved.py#L311

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+                precision=1.0e-6,
+                guess_centroid=None,
+                strict=True,
+                round_moments=False,
+                hsmparams=None,
+                use_sky_coords=False,
+            )
+
+        except galsim.hsm.GalSimHSMError as e:
+            raise MeasurementError(e)
```
arunkannawadi:
```suggestion
            raise MeasurementError(e, self.GALSIM.number)
```
after you transfer the various error flags from the C++ code. Here's a useful document to read: https://github.com/lsst/meas_base/blob/1bfaca56951770c88f4308da41de16d72ce40db9/doc/lsst.meas.base/tasks_and_algorithms.rst#L123

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,208 @@
...
+    def measure(self, record, exposure):
+        """... in place"""
+        # source in c++ is record in py. exposure the same
+
+        center = record.getCentroid()
+        badMaskPlanes = self.config.badMaskPlanes
+        badPixelMask = exposure.getMaskedImage().getMask().getPlaneBitMask(badMaskPlanes)
+        bbox = record.getFootprint().getBBox()
+        if bbox.getArea() == 0:
+            raise MeasurementError("Find the same error message from .cc file")
```
arunkannawadi:
```suggestion
            raise MeasurementError(self.NO_PIXELS.doc, self.NO_PIXELS.number)
```

arunkannawadi: and similarly for other errors.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: It didn't result in an error, but it resulted in nans. Converted them to `Point2D` and fixed it.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: This results in nans. To fix it, I converted `center` to an instance of `galsim.PositionD`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
arunkannawadi: Makes sense.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,243 @@
...
+    def __init__(self, config, name, schema, metadata, logName=None) -> None:
+        super().__init__(config, name, schema, logName)
+        # some flag thingy
+        # work in conjunction with fail()
+        self.centroidResultKey = afwTable.Point2DKey.addFields(
+            schema, name, "The centroid measured by HSM", "pixel"
+        )
+        self.shapeKey = afwTable.QuadrupoleKey.addFields(schema, name, doc="HSM source moments.")
+        if config.addFlux:  # search the correct attribute name in the codebase
+            self.fluxKey = measBase.FluxResultKey.addFields(schema, "flux", "The flux measured by HSM", "dn")
```
arunkannawadi:
```suggestion
            self.fluxKey = measBase.FluxResultKey.addFields(schema, name + "_Flux", "The flux measured by HSM", "dn")
```

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,243 @@
...
+
+class HsmMomentsPlugin(SingleFramePlugin):
+    ConfigClass = HsmMomentsConfig
+
+    def __init__(self, config, name, schema, logName=None):
+        super().__init__(config, name, schema, logName)
+
+        # Flag definitions.
+        flagDefs = FlagDefinitionList()
+        self.FAILURE = flagDefs.add("flag", "General failure flag, set if anything went wrong")
```
arunkannawadi: I suspect that this one alone should be `.addFailureFlag` without the first argument, instead of `.add`. We can come back to this later.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
arunkannawadi: This entire `__init__` method should go to `HsmMomentsPlugin`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,243 @@
...
+
+        if not bbox.contains(Point2I(center)):
+            raise MeasurementError(self.NOT_CONTAINED.doc, self.NOT_CONTAINED.number)
+
+        psfSigma = exposure.getPsf().computeShape(center).getTraceRadius()
+
+        # galsim
+        image_array = exposure[bbox].getImage().array
+        xmin, xmax = bbox.getMinX() + 1, bbox.getMaxX() + 1
+        ymin, ymax = bbox.getMinY() + 1, bbox.getMaxY() + 1
```
arunkannawadi: It looks like that this `+1` is not required, and is perhaps causing the differences between the C++ implementation and pure-Python one.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: I realized the docs need to be class-specific. I kept them in individual classes for now, but we can return to this later.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -88,13 +83,13 @@ def calculate(
                 badpix=badpix,
                 guess_sig=sigma,
                 precision=1.0e-6,
```
arunkannawadi: Make `precision` a config as well in the base class, with `1e-6` as the default value.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,323 @@
...
+        self.flagHandler = measBase.FlagHandler.addFields(schema, name, flagDefs)
+
+        # Safe centroid extractor.
+        self.centroidExtractor = measBase.SafeCentroidExtractor(schema, name)
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.SHAPE_ORDER
+
+    def calculate(
```
arunkannawadi: Make this a 'private' method.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,323 @@
...
+        self,
+        record,
+        *,
+        image: galsim.Image,
+        sigma: float,
+        badpix: galsim.Image | None,
+        centeroid: Point2D,
```
arunkannawadi: Typo. `centroid` everywhere.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,417 @@
...
+    subtractCenter = pexConfig.Field[bool](doc="Subtract starting center from x/y outputs?", default=False)
+
+
+class HsmMomentsPlugin(measBase.SingleFramePlugin):
+    """Base plugin for HSM adaptive moments measurement."""
+
+    ConfigClass = HsmMomentsConfig
+
+    def __init__(self, config, name, schema, metadata, logName=None):
+        super().__init__(config, name, schema, logName)
```
arunkannawadi:
```suggestion
        super().__init__(config, name, schema, metadata, logName=logName)
```

arunkannawadi: The previous signature was incorrect because you were passing `logName` for `metadata`. See the signature of the parent class here: https://github.com/lsst/meas_base/blob/1bfaca56951770c88f4308da41de16d72ce40db9/python/lsst/meas/base/sfm.py#L83

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,417 @@
...
+            Raised for errors in measurement.
+        """
+        # Convert centroid to GalSim's PositionD type.
+        guessCentroid = galsim.PositionD(centroid.x, centroid.y)
+
+        try:
+            # Attempt to compute HSM moments.
+            shape = galsim.hsm.FindAdaptiveMom(
+                image,
+                weight=None,
```
arunkannawadi: The weight needs to be supplied here. It's going to involve implementing the `convertMask` function in Python: https://github.com/lsst/meas_extensions_shapeHSM/blob/main/include/lsst/meas/extensions/shapeHSM/HsmAdapter.h#L82-L100

This would also mean that the method needs to take `weight` as an input argument.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,417 @@
...
+        # Define flags for possible issues that might arise during measurement.
+        flagDefs = measBase.FlagDefinitionList()
+        self.FAILURE = flagDefs.addFailureFlag("General failure flag, set if anything went wrong")
+        self.NO_PIXELS = flagDefs.add("flag_no_pixels", "No pixels to measure")
+        self.NOT_CONTAINED = flagDefs.add(
+            "flag_not_contained", "Center not contained in footprint bounding box"
+        )
+        self.PARENT_SOURCE = flagDefs.add("flag_parent_source", "Parent source, ignored")
+        self.GALSIM = flagDefs.add("flag_galsim", "GalSim failure")
+        self.INVALID_PARAM = flagDefs.add("flag_invalid_param", "Invalid parameter")
```
arunkannawadi:
```suggestion
        self.INVALID_PARAM = flagDefs.add("flag_invalid_param", "Invalid combination of moments")
```

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,417 @@
...
+    )
+
+
+@measBase.register("ext_shapeHSM_HsmSourceMoments")
+class HsmSourceMomentsPlugin(HsmMomentsPlugin):
+    """Plugin for HSM adaptive moments measurement for sources."""
+    ConfigClass = HsmSourceMomentsConfig
+
+    def __init__(self, config, name, schema, metadata, logName=None):
+        super().__init__(config, name, schema, logName)
```
arunkannawadi: Update the signature as above.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,417 @@
...
+        self.centroidResultKey = afwTable.Point2DKey.addFields(schema, name, "HSM source centroid", "pixel")
+        self.shapeKey = afwTable.QuadrupoleKey.addFields(
+            schema, name, "HSM source moments", afwTable.CoordinateType.PIXEL
+        )
+        if config.addFlux:
+            self.fluxKey = measBase.FluxResultKey.addFields(schema, name + "_Flux", "HSM source flux", "dn")
```
arunkannawadi: The units for this quantity is not "dn" for sure. Leave it blank.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,417 @@
...
+        if self.config.addFlux:
+            record.set(self.fluxKey, shape.moments_amp)
+
+        # TODO: calculate errors in shape, centroid?
+
+    def fail(self, record, error=None):
+        # Docstring inherited.
+        self.flagHandler.handleFailure(record)
+        if error:
+            centroid = record.getCentroid()
```
arunkannawadi: Update this to use `centroidExtractor` as well.

arunkannawadi (commented):
The docstrings look fine to me.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Ah, correct! Can we use "count"?
https://github.com/lsst/meas_extensions_trailedSources/blob/3ddfee87b48b9533850c8cd877b2d95ab53c4cf1/python/lsst/meas/extensions/trailedSources/VeresPlugin.py#L110

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: It looks like GalSim already has a python version for it:
https://github.com/GalSim-developers/GalSim/blob/1b56bb4be8602dba9679a0106606b83ca6075609/galsim/hsm.py#L483

enourbakhsh: It seems like the GalSim version I'm looking at takes `weight` and `badpix` as `galsim.Image` objects, and it does the conversion on its own in `galsim.hsm.FindAdaptiveMom()`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Removed it.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,518 @@
...
+
+@measBase.register("ext_shapeHSM_HsmPsfMomentsDebiased")
+class HsmPsfMomentsDebiasedPlugin(HsmPsfMomentsPlugin):
+    """Plugin for debiased HSM adaptive moments measurement for PSFs."""
+
+    ConfigClass = HsmPsfMomentsDebiasedConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.FLUX_ORDER + 1
```
arunkannawadi:
```suggestion
        return cls.FLUX_ORDER + 0.1
```
Since the standard execution order increases in steps of 1, it's safer to keep the increase by hand to less than 1. The exact value does not matter.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,548 @@
...
+
+@measBase.register("ext_shapeHSM_HsmPsfMoments")
+class HsmPsfMomentsPlugin(HsmMomentsPlugin):
+    """Plugin for HSM adaptive moments measurement for PSFs."""
+
+    ConfigClass = HsmPsfMomentsConfig
+
+    def __init__(self, config, name, schema, metadata, logName=None):
+        super().__init__(config, name, schema, metadata, logName=logName)
+        docPrefix = "Debiased centroid" if "Debiased" in self.__class__.__name__ else "Centroid"
```
arunkannawadi: Instead of this, define a private class variable, say `_debiased`, which is `False` for this one, and `True` for the debiased plugin. Condition the `docPrefix` based on that class variable instead of relying on the class name.

arunkannawadi: And if you do that, you could pass that value for `debiasedPsfMoments` in the call to `_measure` and avoid overriding `measure` in `HsmPsfMomentsDebiasedPlugin`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,548 @@
...
+        psfSigma = psf.computeShape(center).getTraceRadius()
+
+        # Get the bounding box in the parent coordinate system.
+        bbox = psfImage.getBBox(afwImage.PARENT)
+
+        # Turn bounding box corners into GalSim bounds.
+        xmin, xmax = bbox.getMinX(), bbox.getMaxX()
+        ymin, ymax = bbox.getMinY(), bbox.getMaxY()
+        bounds = galsim.bounds.BoundsI(xmin, xmax, ymin, ymax)
+
```
arunkannawadi, line 424: For the record, I'm not a big fan of this method doing the debiased version and the regular version based on `debiasedPsfMoments`, but it will do for now.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done! It simplified even more things.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,529 @@
...
+        Parameters
+        ----------
+        record : `~lsst.afw.table.SourceRecord`
+            Record to store measurements.
+        image : `~galsim.Image`
+            Image on which to perform measurements.
+        weight : `~galsim.Image`
+            The weight image for the galaxy being measured. Can be an int or a
+            float array.
+        badpix : `~galsim.Image`
```
jmeyers314: Need some more info here; is zero=good, non-zero=bad?

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,529 @@
...
+import lsst.afw.image as afwImage
+import lsst.afw.math as afwMath
+import lsst.afw.table as afwTable
+import lsst.meas.base as measBase
+import lsst.pex.config as pexConfig
+import numpy as np
+from lsst.geom import Point2D, Point2I
+from lsst.pex.exceptions import InvalidParameterError, NotFoundError
+
+
```
arunkannawadi, line 46: You need an `__all__` list that lists all the configs and plugins, except for the base class ones.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,529 @@
...
+        Calculate adaptive moments using GalSim's HSM and modify the record in
+        place.
+
+        Parameters
+        ----------
+        record : `~lsst.afw.table.SourceRecord`
+            Record to store measurements.
+        image : `~galsim.Image`
+            Image on which to perform measurements.
+        weight : `~galsim.Image`
```
arunkannawadi:
```suggestion
        weight : `~galsim.Image`, optional
```
and set the default to `None` if it's allowed. Similarly for the other ones that accept `None`.  That'd mean you need to move up the `sigma` argument.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,529 @@
...
+            raise measBase.MeasurementError(error, self.INVALID_PARAM.number)
+
+        # Store the quadrupole moments in the record.
+        record.set(self.shapeKey, quad)
+
+        # Store the flux if required by configuration.
+        if self.config.addFlux:
+            record.set(self.fluxKey, shape.moments_amp)
+
+        # TODO: calculate errors in shape, centroid?
```
arunkannawadi: Remove this comment. GalSim doesn't compute these, and we aren't going to either.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
jmeyers314: Yeah, I think we ought to address this now.  One possibility may be to put the logic under the `if self._debiased` block into an overridden method `HsmPsfMomentsDebiasedPlugin.adjust_noise()`, which could just be a noop in `HsmPsfMomentsPlugin`.  That'll keep the logic of debiasing within the Debiased plugin, and I think it'll enable you to get rid of the `_debiased` class variable too.

jmeyers314 (changes requested):
Hi @enourbakhsh , 

  This is really nice!

  I don't have much to add, but I do think it'd be good to address @arunkannawadi 's comment on the debiased logic now.  Feel free to push back if there's some technical reason I'm not appreciating as to why this is hard.

  Lastly, does this PR address anything unearthed by @erykoff 's dive into HSM moments a month ago?  [Here](https://lsstc.slack.com/archives/C2JPMCF5X/p1697156227742329?thread_ts=1697057581.083409&cid=C2JPMCF5X) he wrote:

> It looks like shapeHSM is wrong.  I’m casting a side-eye on the complicated image translation code, which I don’t think is handling the indexing correctly; if I call this in python it gives me an even kernel size! https://github.com/lsst/meas_extensions_shapeHSM/blob/main/include/lsst/meas/extensions/shapeHSM/HsmAdapter.h#L29-L39


If this _does_ change anything related, then it'd be nice to add a unit test that would have caught the error.  (Maybe some of your new tests do address this; I couldn't tell just by reading them).

erykoff:
I haven't looked at the code, but the test that I envisioned is that the plugin should run a gaussian PSF through various boxes, both truncated and not, and running galsim in the tests and meas_extensions_shapeHSM should give identical answers.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Thanks for spotting this. Done!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Added the requested info!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done! I used all the defaults from `galsim` including for `sigma`, so just kept the same order.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Good idea. Implemented!

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -527,3 +479,113 @@ def getExecutionOrder(cls):
...
+            # Scale the noise by the square root of the variance.
+            var.sqrt()  # In-place square root.
+            noise *= var
+
+        # 4. Add the scaled noise to the PSF image.
+        psfImage += noise
+
+        # Masking is needed for debiased PSF moments.
+        badpix = afwImage.Mask(psfBBox)
+        # Safer to uncomment the lines below?
```
enourbakhsh: Question: It appears that the `overlap` obtained from these two lines (currently commented out, but implemented in the C++ version) actually yields the same `overlap` as the one calculated earlier at the beginning of the method. To confirm, should I uncomment these lines to precisely mirror the C++ code, or is it logically redundant to repeat the `overlap` calculation?

enourbakhsh: In other words, given that
```
psf = exposure.getPsf()
psfBBox = psf.computeImageBBox(center)
badpix = afwImage.Mask(psfBBox)
if self.config.useSourceCentroidOffset:
    psfImage = psf.computeImage(center)
else:
    psfImage = psf.computeKernelImage(center)
    psfImage.setXY0(psfBBox.getMin())
```
are these the same?
```
overlap = psfImage.getBBox()
```
and
```
overlap = badpix.getBBox()
```

enourbakhsh:
> Hi @enourbakhsh ,
> 
> This is really nice!
> 
> I don't have much to add, but I do think it'd be good to address @arunkannawadi 's comment on the debiased logic now. Feel free to push back if there's some technical reason I'm not appreciating as to why this is hard.
> 
> Lastly, does this PR address anything unearthed by @erykoff 's dive into HSM moments a month ago? [Here](https://lsstc.slack.com/archives/C2JPMCF5X/p1697156227742329?thread_ts=1697057581.083409&cid=C2JPMCF5X) he wrote:
> 
> > It looks like shapeHSM is wrong.  I’m casting a side-eye on the complicated image translation code, which I don’t think is handling the indexing correctly; if I call this in python it gives me an even kernel size! https://github.com/lsst/meas_extensions_shapeHSM/blob/main/include/lsst/meas/extensions/shapeHSM/HsmAdapter.h#L29-L39
> 
> If this _does_ change anything related, then it'd be nice to add a unit test that would have caught the error. (Maybe some of your new tests do address this; I couldn't tell just by reading them).

Thank you. The plan was to either address the bug in the C++ implementation or utilize GalSim through Python, as it was working for Eli. We decided on the latter approach. I'll make a test to verify that GalSim and `meas_extensions_shapeHSM` give identical answers.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
jmeyers314: I'm only really confident in these being the same for the `True` branch of that `if` statement (though it very well could be true for the `False` branch too). I think I'd recommend uncommenting though.  I think there's zero chance this is a time sink, and using the bbox from the image that's about to be cropped seems safer than using the bbox from a different image, even if they're nominally supposed to have the same bounds.

jmeyers314 (approved):
LGTM.

enourbakhsh:
I added an additional check for severely truncated PSFs, which, as Eli found, exacerbated the problems. Will merge soon if there are no further comments.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Uncommented those two lines.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -0,0 +1,608 @@
...
+
+import galsim
+import lsst.afw.geom as afwGeom
+import lsst.afw.image as afwImage
+import lsst.afw.math as afwMath
+import lsst.afw.table as afwTable
+import lsst.meas.base as measBase
+import lsst.pex.config as pexConfig
+import numpy as np
+from lsst.geom import Point2D, Point2I, Box2I
```
arunkannawadi:
```suggestion
from lsst.geom import Box2I, Point2D, Point2I
```

for it to be compliant with `isort`.

PR #45, opened 2023-11-21
Tickets/DM-41648

ztq1996:
Code for measuring the standardized higher order moments for source and PSF, based on [Zhang et al. (2023)](https://arxiv.org/abs/2212.03257)

The code currently does the second-moment measurement by Galsim itself and uses that to build the weight function and the standardized distance. There is a unit test that ensures the code produces the correct moments for a point source with a Gaussian PSF.


arunkannawadi:
Thanks @ztq1996 . Merging this to the ticket branch to standardize this.

PR #47, opened 2023-12-01
DM-41489: Reimplement C++ HSM shear measurement algorithms in Python

enourbakhsh:


python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+    shearType = pexConfig.ChoiceField[str](
+        doc="The desired method of PSF correction using GalSim. The first "
+        "three options return an e-type distortion, whereas the last option returns a g-type shear.",
+        allowed={
+            "REGAUSS": "From Hirata & Seljak (2003)",
+            "LINEAR": "A modification by Hirata & Seljak (2003) of methods in Bernstein & Jarvis (2002)",
+            "BJ": "From Bernstein & Jarvis (2002)",
+            "KSB": "From Kaiser, Squires, & Broadhurst (1995)",
+        },
+        default="REGAUSS",  # FIXME: double check - it's the default in GalSim though
```
arunkannawadi: The default should be ``REGAUSS`

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+class HsmShapeConfig(measBase.SingleFramePluginConfig):
+    """Base configuration for HSM shape measurement."""
+
+    shearType = pexConfig.ChoiceField[str](
+        doc="The desired method of PSF correction using GalSim. The first "
+        "three options return an e-type distortion, whereas the last option returns a g-type shear.",
+        allowed={
+            "REGAUSS": "From Hirata & Seljak (2003)",
```
arunkannawadi:
```suggestion
            "REGAUSS": "Regaussianization method from Hirata & Seljak (2003)",
```

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+        if self.hasDeblendKey:
+            self.deblendKey = schema[config.deblendNChild]
+
+        self.log = logging.getLogger(self.logName)
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.SHAPE_ORDER
+
+    def addEllipticityField(self, name, n, measType, schema, doc):
```
arunkannawadi: Make this a private static method

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+        # # psfMask = dummyMask
+
+        sctrl = afwMath.StatisticsControl()
+        sctrl.setAndMask(bitValue)
+
+        # NOTE: Origin defaults to PARENT in all cases accessible from Python.
+        variance = afwImage.Image(
+            exposure.variance[bbox],
+            dtype=psfImage.dtype,
+            deep=True,
```
arunkannawadi: Does this have to be a deep copy? Where does it get modified in-place?

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+            "KSB": "From Kaiser, Squires, & Broadhurst (1995)",
+        },
+        default="REGAUSS",  # FIXME: double check - it's the default in GalSim though
+    )
+    measType = pexConfig.ChoiceField[str](
+        doc="Whether to measure ellipticity or shear.",
+        allowed={"ellipticity": "We will measure e1 and e2", "shear": "We will measure g1 and g2"},
+        default="ellipticity",
+        # FIXME: double check (if the default for shearType is REGAUSS then this should also be "ellipticity")
+    )
```
arunkannawadi, line 56: Looking at the code below, I don't think this should be a settable config field at all. The code does not let you run `REGAUSS` with `measType='shear'` and handle the conversion. Suggested refactoring in the comments below.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+        self.GALSIM = flagDefs.add("flag_galsim", "GalSim failure")
+
+        # Embed the flag definitions in the schema using a flag handler.
+        self.flagHandler = measBase.FlagHandler.addFields(schema, name, flagDefs)
+
+        # Utilize a safe centroid extractor that uses the detection footprint
+        # as a fallback if necessary.
+        self.centroidExtractor = measBase.SafeCentroidExtractor(schema, name)
+
+        self.measTypeSymbolLookup = {"ellipticity": "e", "shear": "g"}
```
arunkannawadi: This should be a private class attribute.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+                guess_sig_PSF=psfSigma,
+                precision=1.0e-6,
+                guess_centroid=galsim.PositionD(center.getX(), center.getY()),
+                strict=True,
+                hsmparams=None,
+            )
+        except galsim.hsm.GalSimHSMError as error:
+            raise measBase.MeasurementError(str(error), self.GALSIM.number)
+
+        assert shape.meas_type == self.measTypeSymbolLookup[self.config.measType]
```
arunkannawadi: You should check this and fail with a `FatalAlgorithmError` if these two aren't the same. It indicates that our config is incompatible with `galsim` and there's no point in running it for each object. A `FatalAlgorithmError` would give up on the plugin entirely without wasting time measuring each source and failing every single time.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+        if shape.meas_type == "e":
+            record.set(self.e1Key, shape.corrected_e1)
+            record.set(self.e2Key, shape.corrected_e2)
+            record.set(self.sigmaKey, 2.0 * shape.corrected_shape_err)
+        elif shape.meas_type == "g":
+            record.set(self.e1Key, shape.corrected_g1)
+            record.set(self.e2Key, shape.corrected_g2)
+            record.set(self.sigmaKey, shape.corrected_shape_err)
```
arunkannawadi, line 285: `sigma` here is the size, not the uncertainty estimate. I don't follow why shape errors are assigned to this key here and above.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,364 @@
...
+        else:
+            assert False
```
arunkannawadi: Because `measType` is not user-defined if you make the changes suggested, you can skip the `else` block altogether.

arunkannawadi (dismissed):

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: It's only used in `afwMath.makeStatistics` to get `skyvar`, so at first I thought it'd be safe to use `deep=False` in this case but it gave me this error in the unit tests:

```Traceback (most recent call last):
  File "/home/en7908/repos/meas_extensions_shapeHSM/tests/test_hsm.py", line 420, in testHsmShape
    source = self.runMeasurement(algorithmName, imageid, x_centroid[i], y_centroid[i], sky_var[i])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/en7908/repos/meas_extensions_shapeHSM/tests/test_hsm.py", line 406, in runMeasurement
    plugin.measure(source, exposure)
  File "/home/en7908/repos/meas_extensions_shapeHSM/python/lsst/meas/extensions/shapeHSM/_hsm_shape.py", line 238, in measure
    variance = afwImage.Image(
               ^^^^^^^^^^^^^^^
  File "/scratch/gpfs/HSC/LSST/stack_20230302/conda/envs/lsst-scipipe-7.0.1/share/eups/Linux64/utils/g384e8880d6+81bc2a20b4/python/lsst/utils/wrappers.py", line 303, in __call__
    return clz(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^
lsst.pex.exceptions.wrappers.InvalidParameterError:
  File "include/lsst/afw/image/ImageBase.h", line 219, in lsst::afw::image::ImageBase<PixelT>::ImageBase(const lsst::afw::image::ImageBase<OtherPixelT>&, bool) [with OtherPixelT = float; PixelT = double]
    Only deep copies are permitted for ImageBases with different pixel types {0}
lsst::pex::exceptions::InvalidParameterError: 'Only deep copies are permitted for ImageBases with different pixel types'
```

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: I just followed the C++ code below:
```
   assert(shape.meas_type[0] == measTypeSymbol(_measType));
    switch (*shape.meas_type.c_str()) {
    case 'e':
        source.set(_e1Key, shape.corrected_e1);
        source.set(_e2Key, shape.corrected_e2);
        source.set(_sigmaKey, 2.0*shape.corrected_shape_err);
        break;
    case 'g':
        source.set(_e1Key, shape.corrected_g1);
        source.set(_e2Key, shape.corrected_g2);
        source.set(_sigmaKey, shape.corrected_shape_err);
        break;
    default:
        assert(false);
    }
```
    
Was it a mistake in the C++ implementation? I need to look into it.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Okay, I think I know how to get away with not deep copying here.
```
ipdb> exposure.variance.dtype
<class 'numpy.float32'>
ipdb> psfImage.dtype
<class 'numpy.float64'>
```
It throws an error because we are imposing double precision here while `exposure.variance` is `float32`. The C++ implementation also has `afw::image::Exposure<float>::MaskedImageT::Variance` and not `double`. I changed it to:
```
variance = afwImage.Image(
     exposure.variance[bbox],
     dtype=exposure.variance.dtype,  # NOT psfImage.dtype
     deep=False,
)
```
or even better:
```
variance = afwImage.ImageF(
     exposure.variance[bbox],
     deep=False,
)
```
and it worked!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: I removed `measType` as a configuration option.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: Okay, this is perhaps a documentation error. Reporting `corrected_shape_err` makes sense, but the documentation for this field should read something like "Uncertainty in measured shape (per-component )". I think there would be a similar documentation that you should be able to find from `galsim.ShapeData`.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: On second thoughts, I don't think you should be checking this here at all. This does not depend on the source/object we are measuring and is wasteful to check this for each measurement. We should check that this is true in one of the unit tests and be done with it.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: Nice. I think I like the one where we set `dtype` explicitly based on the `dtype` of `variance` instead of always converting it to an `ImageF`.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,340 @@
...
+        # Embed the flag definitions in the schema using a flag handler.
+        self.flagHandler = measBase.FlagHandler.addFields(schema, name, flagDefs)
+
+        # Utilize a safe centroid extractor that uses the detection footprint
+        # as a fallback if necessary.
+        self.centroidExtractor = measBase.SafeCentroidExtractor(schema, name)
+
+        # Define symbol for measurement type based on shearType.
+        # "g": g-type shear, "e": e-type distortion.
+        self.measTypeSymbol = "g" if config.shearType == "KSB" else "e"
```
arunkannawadi: This is perhaps best set as a class property, where each plugin will set its own `measTypeSymbol` instead of the base class setting it.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,340 @@
...
+
+        self.e1Key = self._addEllipticityField(name, 1, self.measTypeSymbol, schema, self.doc)
+        self.e2Key = self._addEllipticityField(name, 2, self.measTypeSymbol, schema, self.doc)
+        self.sigmaKey = schema.addField(schema.join(name, "sigma"), type=float, doc=f"{self.doc} (width)")
```
arunkannawadi: This docstring is incorrect.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,340 @@
...
+        n : int
+            Specifies whether the field is for the first (1) or second (2)
+            component.
+        measTypeSymbol : MeasType
+            The symbol for measurement type, either 'e' or 'g'.
+        schema : lsst.afw.table.Schema
+            The schema to which the field is added.
+        doc : str
+            The documentation string that needs to be updated to reflect the
+            type and component of the measurement.
```
arunkannawadi, line 139: Since `measTypeSumbol` and `doc` can be fetched from `self`, these can be dropped from the method's signature.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Do you mean something like this?
```
class HsmShapeRegaussConfig(HsmShapeConfig):
    def setDefaults(self):
        super().setDefaults()
        self.shearType = "REGAUSS"
        self.measTypeSymbol = "e"
```
That requires it to be a settable config field, and we decided against it. Maybe you mean this:
```
class HsmShapeRegaussPlugin(HsmShapePlugin):
    ConfigClass = HsmShapeRegaussConfig
    doc = "PSF-corrected shear using Hirata & Seljak (2003) 'regaussianization' method"
    measTypeSymbol = "e"
```

If you mean the latter, doesn't it need to exist in the base class with some default?

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Got it from `galsim.ShapeData`.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Corrected it.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: `measTypeSymbol` won't be accessible from the config. It's not configurable anyway. Just the latter would be good. It's okay for the base class to not have it defined since the base class is not a registered plugin and should not be run by itself anyway.

arunkannawadi:
There's a whole bunch of code deletion that is pending. The folders `include/`, `lib/` and `src/` should be deleted to remove old code. `ups/` should have only the `.table` file and the lines that refer to the `lib` folder should also be removed. The `SConscript` file should be updated to 

```
scripts.BasicSConstruct("meas_extensions_shapeHSM", disableCc=True, noCfgFile=True)
```

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: The problem is that the unit tests do not have access to the `shape` object in order to fetch the `meas_type` from it.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,346 @@
...
+    "HsmShapeRegaussPlugin",
+]
+
+
+class HsmShapeConfig(measBase.SingleFramePluginConfig):
+    """Base configuration for HSM shape measurement."""
+
+    shearType = pexConfig.ChoiceField[str](
+        doc="The desired method of PSF correction using GalSim. The first "
+        "three options return an e-type distortion, whereas the last option returns a g-type shear.",
```
jmeyers314: Similar to Arun's comment, should `shearType` be a config? What happens if you instantiate `HsmShapeBjConfig` with `shearType='LINEAR'`?  No matter what happens, I think this is confusing :)

Could `shearType` simply be a class variable instead?  (That does have the side effect of making HsmShapeConfig purely abstract, so may require an RFC unfortunately).

At the very least, I'd recommend running Config.validate() for each of the subclasses to prohibit examples like above where the class doesn't match the `shearType`.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,346 @@
...
+
+        # Get the PSF image evaluated at the source centroid.
+        psfImage = exposure.getPsf().computeImage(center)
+        psfImage.setXY0(0, 0)
+
+        # Get the trace radius of the PSF.
+        psfSigma = exposure.getPsf().computeShape(center).getTraceRadius()
+
+        # Turn bounding box corners into GalSim bounds.
+        # ipdb> psfImage.getBBox()
```
jmeyers314: Forgot to remove these comments?

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,346 @@
...
+        # ipdb> bbox
+        # Box2I(corner=Point2I(5678, 9876), dimensions=Extent2I(1290, 1274))
+        xmin, xmax = bbox.getMinX(), bbox.getMaxX()
+        ymin, ymax = bbox.getMinY(), bbox.getMaxY()
+        bounds = galsim.bounds.BoundsI(xmin, xmax, ymin, ymax)
+
+        # Each GalSim image below will match whatever dtype the input array is.
+        image = galsim.Image(exposure.image[bbox].array, bounds=bounds, copy=False)
+        psf = galsim.Image(psfImage.array, copy=False)
+        # FIXME: no bounds in the line above? apparently no bounds in the C++ version
```
jmeyers314: PSF is already restricted to a small image, no?  So I think no bounds for the PSF is expected (If that's what this FIXME refers to)

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,346 @@
...
+
+        # Get the `lsst.meas.base` mask for bad pixels.
+        subMask = exposure.mask[bbox]
+        # FIXME: is the above PARENT? we want subMask(*afwMask, bbox, afw::image::PARENT);
```
jmeyers314: above looks right to me.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,346 @@
...
+            )
+
+
+class HsmShapeBjConfig(HsmShapeConfig):
+    """Configuration for HSM shape measurement for the BJ estimator."""
+
+    def setDefaults(self):
+        super().setDefaults()
+        self.shearType = "BJ"
+
```
jmeyers314, line 310: I'd add:

```python
def validate(self):
    super().validate()
    if self.shearType != "BJ":
        raise measBase.FatalAlgorithmError("Incompatible shearType")
```

(or something like that; maybe there's a better exception class to use)...

and similar down below.

jmeyers314 (changes requested):
Looks great @enourbakhsh !  I left you a few questions/comments.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: Yeah, that would break running newer versions of the stack if you try running with older version of the configs. While that is never advisable or guaranteed, we'd want to not break it if we can. I'd support going with a `validate` method that checks this though.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: The unit test could call `FindAdaptiveMom` to get a `ShapeData` object and check this.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
arunkannawadi: Yeah, and it's a good practice to call the `super().validate()` at the end of the method rather at the beginning. Errors thrown from `validate` method should also be  of the type`FieldValidationError` .

arunkannawadi:
You can already follow along the changes being made in #48 and at least use `galsim._Image` and `galsim._Bounds` instances.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: I believe you're referring to `EstimateShear()`. Initially, I tried to avoid it due to the numerous variables required for the call. However, I eventually implemented a minimal version in the unit test, which suffices to determine the measurement type outputted by GalSim after the shear estimation process.

tests/test_hsm.py
```diff
@@ -405,7 +405,31 @@ def runMeasurement(self, algorithmName, imageid, x, y, v):
...
+            PSF_image=psf,
+            shear_est=control.shearType,
+            recompute_flux="FIT",
+            guess_sig_gal=2.5 * psfSigma,
+            guess_sig_PSF=psfSigma,
+            precision=1.0e-6,
+            guess_centroid=galsim.PositionD(center.getX(), center.getY()),
+            strict=True,
+            hsmparams=None,
+        ).meas_type
```
enourbakhsh: Should we make it even more minimal?

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Done!

tests/test_hsm.py
arunkannawadi: What do you have in mind here? Unless the values passed here are the default values for `EstimateShear`, I'd be a bit skeptical about it.

enourbakhsh: This is regarding your comment [here](https://github.com/lsst/meas_extensions_shapeHSM/pull/47#discussion_r1413199861), where you noted that it is not necessary to verify if the `meas_type` output by GalSim aligns with our assumptions for each measurement within the plugin. You suggested we validate this in one of the unit tests. The ones that are not passed will be set to the default of `EstimateShear`. We are not passing `badpix` and `sky_var` because obtaining them here is unnecessary, as they do not affect the value we aim to extract, which is simply `mea_stype`.

enourbakhsh: From GalSim:
``` python
def EstimateShear(gal_image, PSF_image, weight=None, badpix=None, sky_var=0.0,
                  shear_est="REGAUSS", recompute_flux="FIT", guess_sig_gal=5.0,
                  guess_sig_PSF=3.0, precision=1.0e-6, guess_centroid=None,
                  strict=True, hsmparams=None):
                  ...
```

```
After a measurement is made, the type of shape measurement is stored in the
ShapeData structure as 'meas_type' (a string that equals either 'e' or 'g').
```

enourbakhsh: So, we can do:
``` python
        postEstimationMeasType = galsim.hsm.EstimateShear(
            gal_image=image,
            PSF_image=psf,
            shear_est=control.shearType,
            guess_sig_gal=2.5 * psfSigma,
            guess_sig_PSF=psfSigma,
            guess_centroid=galsim.PositionD(center.getX(), center.getY()),
        ).meas_type
```

README.rst
```diff
@@ -9,6 +9,6 @@ The algorithm was initially described in https://ui.adsabs.harvard.edu/abs/2003M
 and was modified later in https://ui.adsabs.harvard.edu/abs/2005MNRAS.361.1287M/abstract.
 HSM is named after the primary authors: i) Hirata, Christopher, ii) Seljak, Uros and iii) Mandelbaum, Rachel.
 Their implementation of this algorithm lives within `GalSim <https://github.com/GalSim-developers/GalSim>`_,
-and this package interacts directly with the C++ layer of GalSim to make the measurements.
+and this package interacts with the Python layer of GalSim to make the measurements.
```
arunkannawadi, line 12: Add a note that we use many `GalSim` primitives that removes the overhead by avoiding many sanity checks.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,372 @@
...
+        # Get the bounding box of the PSF in the parent coordinate system.
+        psfBBox = psfImage.getBBox(afwImage.PARENT)
+
+        # Turn the PSF bounding box corners into GalSim bounds.
+        psfBounds = self.bboxToGalSimBounds(psfBBox)
+
+        # Each GalSim image below will match whatever dtype the input array is.
+        # NOTE: PSF is already restricted to a small image, so no bbox for
+        # the PSF is expected.
+        image = galsim._Image(exposure.image[bbox].array, bounds, None)
```
arunkannawadi:
```suggestion
        image = galsim._Image(exposure.image[bbox].array, bounds, wcs=None)
```

and in other places.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,372 @@
...
+            exposure.variance[bbox],
+            dtype=exposure.variance.dtype,
+            deep=False,
+        )
+        stat = afwMath.makeStatistics(variance, subMask, afwMath.MEDIAN, sctrl)
+
+        skyvar = stat.getValue(afwMath.MEDIAN)
+
+        try:
+            shape = galsim.hsm.EstimateShear(
```
arunkannawadi: I bet this is slow as well and need to call how `_FindAdaptiveMom` was called in `_hsm_moments,py`.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Done!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Done!

tests/test_hsm.py
arunkannawadi: Yeah, you could remove all the guess parameters and pass `strict=False`. The `meas_type` should be the same regardless of the guesses.

enourbakhsh: Sounds good, except I still need to pass `guess_centroid`, otherwise it returns `None` for `meas_type` with `strict=False` and throws this error with `strict=True`:
```
galsim.errors.GalSimHSMError: Bounds don't make sense
````

File: README.rst [hunk 22]
```diff
@@ -9,6 +9,6 @@ The algorithm was initially described in https://ui.adsabs.harvard.edu/abs/2003M
 and was modified later in https://ui.adsabs.harvard.edu/abs/2005MNRAS.361.1287M/abstract.
 HSM is named after the primary authors: i) Hirata, Christopher, ii) Seljak, Uros and iii) Mandelbaum, Rachel.
 Their implementation of this algorithm lives within `GalSim <https://github.com/GalSim-developers/GalSim>`_,
-and this package interacts directly with the C++ layer of GalSim to make the measurements.
+and this package interacts with the Python layer of GalSim to make the measurements. We use several GalSim primitives to reduce overhead by bypassing extensive sanity checks.
```
jmeyers314: Please keep a consistent line length here.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,391 @@
...
+            # Initialize an instance of ShapeData to store the results.
+            shape = galsim.hsm.ShapeData(
+                image_bounds=galsim._BoundsI(0, 0, 1, 1),
+                observed_shape=galsim._Shear(0j),
+                psf_shape=galsim._Shear(0j),
+                moments_centroid=galsim._PositionD(0, 0),
+            )
+
+            # Prepare various values for GalSim's EstimateShearView.
+            recomputFlux = "FIT"
```
jmeyers314:
```suggestion
            recomputeFlux = "FIT"
```

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -0,0 +1,391 @@
...
+            # TODO: [DM-42047] Change to public API when an optimize version is
+            # available.
+            galsim._galsim.EstimateShearView(
+                shape._data,  # shape data buffer (not passed in pure Python)
+                image._image,  # gal_image
+                psf._image,  # PSF_image
+                weight._image,  # weight
+                float(skyvar),  # sky_var
+                self.config.shearType.upper(),  # shear_est
+                recomputFlux.upper(),  # recompute_flux
```
jmeyers314:
```suggestion
                recomputeFlux.upper(),  # recompute_flux
```

tests/test_hsm.py
jmeyers314, line 1: Can you add a test that the new child class `.validate()` methods are functioning?

jmeyers314 (approved):
LGTM!  Left you a few small items.

README.rst
arunkannawadi: And each sentence should be its own line as well.

enourbakhsh: Done!

tests/test_hsm.py
enourbakhsh: Done!

arunkannawadi (approved):

PR #48, opened 2023-12-04
DM-41994: Use galsim primitives vs higher level interface.

erykoff:


python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -91,12 +108,11 @@ def _calculate(
...
         *,
-        image: galsim.Image,
-        weight: galsim.Image | None = None,
-        badpix: galsim.Image | None = None,
+        image: galsim._galsim.ImageViewD | galsim._galsim.ImageViewF,
+        weight_image: galsim._galsim.ImageViewI,
+        centroid: Point2D,
         sigma: float = 5.0,
         precision: float = 1.0e-6,
-        centroid: Point2D | None = None,
```
arunkannawadi, line 99: Good catch. This routine doesn't really support `centroid` being None.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -130,21 +140,25 @@ def _calculate(
...
-                precision=precision,
-                guess_centroid=guessCentroid,
-                strict=True,
-                round_moments=self.config.roundMoments,
-                hsmparams=None,
+                weight_image,
+                float(sigma),
+                float(precision),
+                guessCentroid._p,
+                bool(self.config.roundMoments),
```
arunkannawadi, line 147: Are casting necessary here, especially this one?

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
erykoff: It didn't before, and it still doesn't.  (It could be added, but I don't think that's necessary).

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
erykoff: The casting is just-in-case to make sure that we get the precise datatype match to what the c++ expects.  In these cases I always cast to avoid any ... unpleasant surprises.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -44,6 +44,23 @@
...
+            gcls = galsim._galsim.ImageViewF
+        case np.int32:
+            gcls = galsim._galsim.ImageViewI
+
+    return gcls(
+        array.__array_interface__["data"][0],
+        array.strides[1]//array.itemsize,
+        array.strides[0]//array.itemsize,
+        bounds._b,
+    )
```
rmjarvis: Is this actually faster than using `galsim._Image(array, bounds, None)`?  That's the supported API that skips most of the sanity checks.  This interface isn't guaranteed to be stable across versions.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -130,21 +140,25 @@ def _calculate(
...
-
         try:
             # Attempt to compute HSM moments.
-            shape = galsim.hsm.FindAdaptiveMom(
+
+            # Use galsim c++/python interface directly.
+            shape = galsim.hsm.ShapeData()
+            hsmparams = galsim.hsm.HSMParams.default
+
+            galsim._galsim.FindAdaptiveMomView(
```
rmjarvis, line 140: Even this isn't really guaranteed to be stable, but it should be very close to equivalent to `image.FindAdaptiveMom(...)`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -278,19 +292,20 @@ def measure(self, record, exposure):
         # Create a GalSim image using the extracted array.
         # NOTE: GalSim's HSM uses the FITS convention of 1,1 for the
         # lower-left corner.
-        image = galsim.Image(imageArray, bounds=bounds, copy=False)
+        _image = _quickGalsimImageView(imageArray, bounds)
```
rmjarvis: Here, I would recommend `image = galsim._Image(imageArray, bounds, None)`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -430,7 +445,17 @@ def measure(self, record, exposure):
         imageArray = psfImage.array
 
         # Create a GalSim image using the PSF image array.
-        image = galsim.Image(imageArray, bounds=bounds, copy=False)
+        _image = _quickGalsimImageView(imageArray, bounds)
```
rmjarvis: Also here.  `image = galsim._Image(imageArray, bounds, None)`.

rmjarvis (commented):
I wouldn't recommend going this way.  This is using implementation details in ways that are basically equivalent to the C++ syntax reliance that you currently have in the stack.  I assume most of the reason for this is to avoid the heavy overhead in the `galsim.Image` constructor.  

However, you should be able to remove most of that using the supported `galsim._Image` factory function.  It's a leading underscore function, but it is documented, which means it is guaranteed to be API stable using the normal semantic versioning rules.  cf. http://galsim-developers.github.io/GalSim/_build/html/image_class.html#galsim._Image

I think using that would get you most of the performance you're looking for here without the fragility of relying on implementation details.

arunkannawadi:
@rmjarvis recommends using galsim._Image to avoid the overhead instead of the APIs that aren’t guaranteed to be stable. In hindsight, I should have known it but completely forgot about it. That should be the best alternative to _quickGalsimImageView. We don’t gain much by using FindAdaptiveMomView directly so perhaps we should stick to using the Python interface there.

http://galsim-developers.github.io/GalSim/_build/html/image_class.html#galsim._Image

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
erykoff: Thanks!  I didn't know about this, and will take a look.  It does look like it should do what we need, and will confirm.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
erykoff: So there was actually a significant amount of time spent in `_convertMask` which was some of the motivation to use the primitive.  If there was an officially supported way of bypassing this I think we could use the official API.

arunkannawadi: Oh right. The direct Python API calls the `_convertMask` [function ](https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/hsm.py#L483-L529) which has a bunch of `galsim.Image` instantiations making it slow as well. I could open an issue on the GalSim repo if there's no immediate supported API.

rmjarvis: That's fair.  What mask do you use?  None?  Or something real?

arunkannawadi: We supply a value for `badpix` that masks some pixels, so very much real. I think the overheads there can be reduced by using `galsim._ImageI` instead of `galsim.ImageI` in `_convertMask`.

arunkannawadi: https://github.com/GalSim-developers/GalSim/issues/1264

arunkannawadi:
Eli, for this ticket, should we just move on to `galsim._Image` and perhaps `galsim._Bounds` and close this ticket? I suspect it'll be a while before a more supported API for `FindAdaptiveMom` will make its way to rubin-env anyway.

erykoff:
@arunkannawadi This is ready for re-review.  I have done what you suggest, and I didn't know about the `_BoundsI` constructor until now.

arunkannawadi:
That was also a suggestion from Mike on Slack (DM). Per my calculations, it should save us some 5% of the total run time on single-visit images (and lesser on coadds).

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -130,21 +123,25 @@ def _calculate(
...
-                badpix=badpix,
-                guess_sig=sigma,
-                precision=precision,
-                guess_centroid=guessCentroid,
-                strict=True,
-                round_moments=self.config.roundMoments,
-                hsmparams=None,
+
+            # Use galsim c++/python interface directly.
+            shape = galsim.hsm.ShapeData()
```
arunkannawadi: Pass in arguments here: `observed_shape=galsim._Shear(), psf_shape=galsim._Shear(), moments_centroid=galsim._PositionD()`. These avoid initializing objects underneath without the overhead.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -130,21 +123,25 @@ def _calculate(
         """
         # Convert centroid to GalSim's PositionD type.
         guessCentroid = galsim.PositionD(centroid.x, centroid.y)
```
arunkannawadi: Make this `_PositionD`

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
arunkannawadi: Right now, this is the only sticking point I believe. Either we use the `FindAdaptiveMom` call or have a TODO comment here referring to a ticket that should replace it with `FindAdaptiveMom` as soon as we have a version of GalSim with the above issue implemented.

arunkannawadi (approved):

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -165,9 +169,10 @@ def _calculate(
         # Convert GalSim measurements to lsst measurements.
         try:
             # Create an ellipse for the shape.
+            observed_shape = shape.observed_shape
             ellipse = afwGeom.ellipses.SeparableDistortionDeterminantRadius(
-                e1=shape.observed_shape.e1,
-                e2=shape.observed_shape.e2,
+                e1=observed_shape.e1,
+                e2=observed_shape.e2,
```
arunkannawadi, line 175: Capturing the observed_shape in a variable is great, because getting this attribute is expensive. It constructs a `Shear` object, with overheads, each time `observed_shape` is accessed from a `ShapeData` object.

PR #49, opened 2023-12-15
DM-42170: Add unit test to verify shapeHSM shear measurements against GalSim's Python layer outputs

enourbakhsh:


tests/test_hsm.py
```diff
@@ -507,6 +584,105 @@ def testHsmShape(self):
...
+                # Increasing the width beyond 4.5 leads to noticeable
+                # truncation of the PSF, i.e. a PSF that is too large for the
+                # box. While this truncated state leads to incorrect
+                # measurements, it is necessary for testing purposes to
+                # evaluate the behavior under these extreme conditions.
+                # Increasing the width beyond 41.3 fails to converge for this
+                # particular test dataset.
+                width=(2.0, 3.0, 4.0, 10.0, 40.0),
+                varyBBox=(True, False),
+                wrongBBox=(True, False),
```
enourbakhsh: The test fails for runs with `wrongBBox=True` because the output ellipticities are not identical, e.g.
```
AssertionError: -2.957377418855763e-16 != -2.9524050707575844e-16
```

erykoff: Looking at the test, I think that’s using the incorrect bbox. Eg the galsim test may be wrong and not the code.

enourbakhsh: If I replace
```python
psf = galsim.Image(psfImage.array, copy=False)
```
with
```python
psfBBox = psfImage.getBBox(afwImage.PARENT)
psfBounds = galsim.bounds.BoundsI(psfBBox.getMinX(), psfBBox.getMaxX(), psfBBox.getMinY(), psfBBox.getMaxY())
psf = galsim.Image(psfImage.array, bounds=psfBounds, copy=False)
```
in `computeDirectShapeFromGalSim()`, it solves the problem!

tests/test_hsm.py
```diff
@@ -727,7 +727,7 @@ def _doComputeImage(self, position=None, color=None):
             # inconsistent.  Old shapeHSM code attempted to infer the former
             # from the latter, but was unreliable.  New code infers the former
             # directly, so this inconsistency no longer breaks things.
-            bbox.shift(geom.Extent2I(1, 1))
+            bbox.shift(geom.Extent2I(1, 2))
```
enourbakhsh, line 723: Here, I've adjusted the shift from `Extent2I(1, 1)` to `Extent2I(1, 2)` to reduce confusion during debugging. This modification aids in distinguishing that some errors are not just related to the 0 vs 1 indexing issue.

arunkannawadi: This is good. It's always better to have different values in `x` and `y` to catch any errors that may coincidentally cancel each other.

tests/test_hsm.py
```diff
@@ -366,6 +366,89 @@ def tearDown(self):
...
+        # Get the center of the source as a Point2D.
+        center = geom.Point2D(record.get("centroid_x"), record.get("centroid_y"))
+
+        # Get the PSF image evaluated at the source centroid.
+        psfImage = exposure.getPsf().computeImage(center)
+        psfImage.setXY0(0, 0)
+
+        # Get the GalSim images to use in the EstimateShear call.
+        bbox = record.getFootprint().getBBox()
+        bounds = galsim.bounds.BoundsI(bbox.getMinX(), bbox.getMaxX(), bbox.getMinY(), bbox.getMaxY())
```
arunkannawadi: This, and other instances, can be just `galsim.BoundsI`.

tests/test_hsm.py
```diff
@@ -366,6 +366,89 @@ def tearDown(self):
...
+        image = galsim.Image(exposure.image[bbox].array, bounds=bounds, copy=False)
+        psfBBox = psfImage.getBBox(afwImage.PARENT)
+        psfBounds = galsim.bounds.BoundsI(
+            psfBBox.getMinX(), psfBBox.getMaxX(), psfBBox.getMinY(), psfBBox.getMaxY()
+        )
+        psf = galsim.Image(psfImage.array, bounds=psfBounds, copy=False)
```
arunkannawadi: To make this test a bit more non-trivial, remove `copy=False` everywhere. On the off chance if something that shouldn't change in-place changes, we could catch it.

tests/test_hsm.py
```diff
@@ -366,6 +366,89 @@ def tearDown(self):
...
+        badpix &= bitValue
+        badpix = galsim.Image(badpix, bounds=bounds, copy=False)
+
+        # Estimate the sky variance.
+        sctrl = afwMath.StatisticsControl()
+        sctrl.setAndMask(bitValue)
+        variance = afwImage.Image(
+            exposure.variance[bbox],
+            dtype=exposure.variance.dtype,
+            deep=False,
```
arunkannawadi: Set `deep=True` just as you set `copy=True` (or remove it) to create `galsim.Image` instances.

tests/test_hsm.py
```diff
@@ -507,6 +590,101 @@ def testHsmShape(self):
 
         self.assertEqual(nFail, 0, "\n"+msg)
 
+    @lsst.utils.tests.methodParameters(
```
arunkannawadi: Use `methodParametersProduct` instead of `methodParameters`. Then you wouldn't have to do the `dict` of `list` of `zip` etc.

arunkannawadi (approved):

tests/test_hsm.py
enourbakhsh: Done!

tests/test_hsm.py
enourbakhsh: That's really convenient. Done!

tests/test_hsm.py
enourbakhsh: Done!

tests/test_hsm.py
enourbakhsh: Done!

tests/test_hsm.py
```diff
@@ -366,6 +366,89 @@ def tearDown(self):
...
+
+        Parameters
+        ----------
+        record : `~lsst.afw.table.SourceRecord`
+            The record containing the center and footprint of the source which
+            needs measurement.
+        exposure : `~lsst.afw.image.Exposure`
+            The exposure containing the source which needs measurement.
+        config : `~lsst.meas.extensions.shapeHSM._hsm_shape.\
+                 HsmShape<Method>Config`
```
arunkannawadi:
```suggestion
                 HsmShapeConfig`
```
This will include all its subclasses. You can specify that this has to be one of the child classes in the description that follows.

tests/test_hsm.py
```diff
@@ -366,6 +366,89 @@ def tearDown(self):
...
+            The configuration object containing parameters and settings for
+            this measurement. <Method> is the name of the correction method
+            being used (e.g., Ksb, Regauss, etc.).
+
+        Returns
+        -------
+        `~galsim.hsm.ShapeData`
```
arunkannawadi:
```suggestion
        shape: `~galsim.hsm.ShapeData`
```

arunkannawadi: The variable name doesn't matter, but one has to be specified.

arunkannawadi (approved):

PR #68, opened 2024-02-17
DM-42294: Verify shapeHSM and GalSim consistency across different zero-padding levels on truncated PSFs

enourbakhsh:


erykoff (approved):
Please check the black formatting and this is good to go.

PR #70, opened 2024-04-24
DM-43970: Fix unexpected errors in HsmShapeRegauss reported as warnings

enourbakhsh:


python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -259,20 +259,25 @@ def measure(self, record, exposure):
...
-                float(skyvar),  # sky_var
-                self.config.shearType.upper(),  # shear_est
-                recomputeFlux.upper(),  # recompute_flux
-                float(2.5 * psfSigma),  # guess_sig_gal
-                float(psfSigma),  # guess_sig_PSF
-                float(precision),  # precision
-                guessCentroid._p,  # guess_centroid
-                hsmparams._hsmp,  # hsmparams
-            )
+            try:
```
arunkannawadi: You shouldn't need a `try` block within a `try` block. Just catch both types of errors in the same `except` block below. Reference for capturing two types of errors: https://docs.python.org/3/tutorial/errors.html#handling-exceptions

enourbakhsh: Oh, you're right, that's way better!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -109,7 +109,7 @@ def __init__(self, config, name, schema, metadata, logName=None):
         self.hasDeblendKey = len(config.deblendNChild) > 0
 
         if self.hasDeblendKey:
-            self.deblendKey = schema[config.deblendNChild]
+            self.deblendKey = schema.find(config.deblendNChild).key
```
enourbakhsh: Is this `schema.find` approach considered standard?

arunkannawadi: Was there a problem with how it was prior to the change?

enourbakhsh: Yes, it was causing `'lsst.afw.table._table.SubSchema' object has no attribute 'get'` with `self.hasDeblendKey==True`.

parejkoj: Yes, this was incorrect before. Either what @enourbakhsh did here, or `schema[name].asKey()` is also equivalent.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -273,7 +273,9 @@ def measure(self, record, exposure):
                 guessCentroid._p,  # guess_centroid
                 hsmparams._hsmp,  # hsmparams
             )
-        except galsim.hsm.GalSimHSMError as error:
+        # GalSim does not raise custom pybind errors as of version 2.5,
+        # resulting in all GalSim C++ errors being RuntimeErrors.
+        except (galsim.hsm.GalSimHSMError, RuntimeError) as error:
```
parejkoj: Is there any concern that these RuntimeError's raised by GalSim might be more serious ones that we should further investigate, or can we assume that they're all just internal algorithmic problems and ignore them (which is what this is effectively doing)?

parejkoj (approved):
Looks good. One question about GalSim which might help us determine if we need to file other tickets.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
taranu: The errors were mostly being thrown on objects that were flagged but not ignored due to the bug above, so I'm not too worried about that. They are unfortunately not granular so if you wanted to add more specific flags you'd need to parse the error message.

Note that the comment above mentions DM-42047, which will change the line above to call GalSim's Python interface. But I see that [that interface](https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/hsm.py#L700) just catches `RuntimeError` and re-raises as `GalSimHSMError`, which will end up in the same result.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
taranu: The point of wrapping only that one line in another try block was that the `RuntimeErrors` were guaranteed to come from GalSim C++. That's still probably going to be true for the rest of the block but not necessarily if someone adds other bits of non-GalSim code to it.

enourbakhsh: The rest of the code in the block doesn't seem troublesome to me, but I agree with @taranu that keeping them there poses the risk of others adding non-GalSim code to it. Looking at the [original C++ code](https://github.com/lsst/meas_extensions_shapeHSM/blob/87969a01f44e02d64a0745474a40150de2db046b/src/HsmShape.cc#L191) I used for this Python implementation, it actually makes more sense to use one `try...except` block and only include GalSim's `EstimateShearView` call in this block. This way, catching both `GalSimHSMError` and `RuntimeError` with this single block addresses Dan's concerns.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: The `schema.find(name).key` method is thanks to Dan's input. If they're both equivalent, I personally find the `schema[name].asKey()` method easier to read/digest.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
taranu: That makes sense. The rest of the block is setup code that should not fail and if it does, it's likely a usage error or something we should catch earlier (i.e. a zero-sized image or something like that), not a GalSim/HSM algorithmic failure.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: I also agree that we can assume these are internal GalSim issues that are expected for some objects and don't need further investigation from us. Additionally, parsing the error message and creating more flags doesn't seem future-proof to me.

PR #72, opened 2024-05-21
DM-43894: Avoid expected warnings in HSM higher order moments plugin

arunkannawadi:


python/lsst/meas/extensions/shapeHSM/_hsm_higher_moments.py
```diff
@@ -369,9 +369,13 @@ def measure(self, record, exposure):
         except KeyError:
             raise measBase.FatalAlgorithmError("'ext_shapeHSM_HsmPsfMoments' plugin must be enabled.")
 
+        centroid = self.centroidExtractor(record, self.flagHandler)
+
+        if record["ext_shapeHSM_HsmPsfMoments_flag"]:
+            raise measBase.MeasurementError(self.FAILURE.doc, self.FAILURE.number)
+
```
leeskelvin: Does the record flag require the `centroid` parameter to be set? I'm not quite sure why you decided to move it above the raise here.

python/lsst/meas/extensions/shapeHSM/_hsm_higher_moments.py
```diff
@@ -369,9 +369,13 @@ def measure(self, record, exposure):
         except KeyError:
             raise measBase.FatalAlgorithmError("'ext_shapeHSM_HsmPsfMoments' plugin must be enabled.")
 
+        centroid = self.centroidExtractor(record, self.flagHandler)
+
+        if record["ext_shapeHSM_HsmPsfMoments_flag"]:
```
leeskelvin: Would an in-line comment above this `if` help explain the rationale for this new addition?

leeskelvin (approved):

python/lsst/meas/extensions/shapeHSM/_hsm_higher_moments.py
arunkannawadi: Ah, nice catch, Lee! I meant to log (at the DEBUG level) that a measurement at this location failed. I realized that that gets logged anyway in the measurement framework here: https://github.com/lsst/meas_base/blob/main/python/lsst/meas/base/baseMeasurement.py#L407-L409

So I decided to skip logging here. I'll move the `centroid` back to where it was.

python/lsst/meas/extensions/shapeHSM/_hsm_higher_moments.py
```diff
@@ -361,15 +366,25 @@ def __init__(self, config, name, schema, metadata, logName=None):
 
     def measure(self, record, exposure):
         # Docstring inherited.
-        M = np.zeros((2, 2))
+
+        # If the PSF model is not available, this plugin would fail for all entries.
+        if (psf := exposure.getPsf()) is None:
+            raise measBase.FatalAlgorithmError("No PSF attached to the exposure.")
```
arunkannawadi, line 372: If the `HsmPsfMoments` algorithm failed because of lack of PSF, I'd like this plugin to also fail fatally instead of attempting every single source/object and failing individually. So I moved this up here.

arunkannawadi (commented):
Lee, I changed more lines than I set out to. Would you mind taking a quick look again? I'll run Jenkins again before merging.

leeskelvin:
Thanks Arun! I like the reformat here, which reads more cleanly to me than it was prior. I think the changes here make sense, and thank you for adding some in-line comments to help with that. I have nothing to add, LGTM!

PS - :crossed_fingers: for Jenkins

leeskelvin (approved):

PR #73, opened 2024-08-07
DM-42282: Fix how FieldValidationError is raised in shapeHSM validate calls and add relevant tests

enourbakhsh:


arunkannawadi (approved):
LGTM! A couple of minor suggestions:

1. Split the additions to the unit test as its own commit in the end.
2. It might be worth adding a class attribute (`SHEAR_TYPE` or something) that holds the value `KSB`, `REGAUSS` etc. The `setDefaults` can set the default `shearType` from the class variable and the `validate` method can check that `shearType` matches the class attribute. This would remove both the `setDefaults` and `validate` methods from children classes and have them both in just the parent class.

enourbakhsh:
> LGTM! A couple of minor suggestions:
> 
> 1. Split the additions to the unit test as its own commit in the end.
> 2. It might be worth adding a class attribute (`SHEAR_TYPE` or something) that holds the value `KSB`, `REGAUSS` etc. The `setDefaults` can set the default `shearType` from the class variable and the `validate` method can check that `shearType` matches the class attribute. This would remove both the `setDefaults` and `validate` methods from children classes and have them both in just the parent class.

Thanks for the review. I did number 1 but filed a new (non-bug) ticket for number 2 since it requires structural changes and additional considerations beyond fixing calls. Also, I want to make it more robust.

PR #74, opened 2024-08-08
DM-45662: Use non-settable shear type to simplify HsmShapeConfig and prevent user errors

enourbakhsh:


python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -67,6 +67,28 @@ class HsmShapeConfig(measBase.SingleFramePluginConfig):
         default=["BAD", "SAT"],
     )
 
+    @property
+    def SHEAR_TYPE(self):
```
arunkannawadi: Define this as `shearType`.

arunkannawadi (changes requested):
If you're going the property route, we can make this much much cleaner.

Remove `shearType` from the config field (without deprecating, yes really) and instead rename `SHEAR_TYPE` as `shearType`.  This quacks like a config, but is not settable like a `pex_config` field. Then, the overrides for `setDefaults` and `validate` can be removed even in the base class (essentially undoing your previous ticket).

To maintain backwards compatibility, you need to define a setter but that shouldn't do anything other than emitting a deprecation warning, with deprecating timeline as in the dev guide. Since this is not a config, you'll need to decorate it with a `deprecated` decorator: https://developer.lsst.io/stack/deprecating-interfaces.html#python-deprecation

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -308,18 +330,9 @@ def fail(self, record, error=None):
...
-    def validate(self):
-        if self.shearType != "BJ":
-            raise pexConfig.FieldValidationError(
-                self.__class__.shearType,
-                self,
-                f"shearType '{self.shearType}' is not valid. It must be 'BJ'.",
-            )
-        super().validate()
+    @property
+    def SHEAR_TYPE(self):
```
arunkannawadi:
```suggestion
    def SHEAR_TYPE(self):
        # Docstring inherited.
```

enourbakhsh:
> If you're going the property route, we can make this much much cleaner.
> 
> Remove `shearType` from the config field (without deprecating, yes really) and instead rename `SHEAR_TYPE` as `shearType`. This quacks like a config, but is not settable like a `pex_config` field. Then, the overrides for `setDefaults` and `validate` can be removed even in the base class (essentially undoing your previous ticket).
> 
> To maintain backwards compatibility, you need to define a setter but that shouldn't do anything other than emitting a deprecation warning, with deprecating timeline as in the dev guide. Since this is not a config, you'll need to decorate it with a `deprecated` decorator: https://developer.lsst.io/stack/deprecating-interfaces.html#python-deprecation

Good idea, I wasn't sure if it was allowed, so I'm glad to hear that. It seems that python does not inherit the setter from the base class property if you override that property in subclasses, which causes an `AttributeError` when setting the property. This meant I had to define a setter for each subclass. However, I realized you can use `@HsmShapeConfig.shearType.getter` to basically create a copy of the property with a different getter that doesn’t wash out the no-op setter defined in the base class. I’ve set up a code removal ticket for this, which is blocked by v28 and will block v29.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -67,6 +54,37 @@ class HsmShapeConfig(measBase.SingleFramePluginConfig):
...
+          Bernstein & Jarvis (2002).
+        - "BJ": The method developed by Bernstein & Jarvis (2002).
+        - "KSB": The method from Kaiser, Squires, & Broadhurst (1995).
+
+        The default method is "REGAUSS". Subclasses can override this property,
+        but it cannot be set externally, making it effectively read-only. The
+        setter is deprecated and will not do anything other than emit a
+        deprecation warning.
+        """
+        return "REGAUSS"
```
arunkannawadi: Raise `NotImplementedError` here saying that this must be implemented in the subclasses.

arunkannawadi:
This way of defining `getter`s is new to me as well. Could you add a unit test that tests that tests that you get the expected `shearType` values when all these plugins are registered simultaneously? An alternative is to define `_shearType` in the child classes and have the base class return `_shearType` if you access `shearType` (in which case, my comment about `NotImplementedError` is invalid). This is not any less read-only way to go about it.

enourbakhsh:
> This way of defining `getter`s is new to me as well. Could you add a unit test that tests that tests that you get the expected `shearType` values when all these plugins are registered simultaneously? An alternative is to define `_shearType` in the child classes and have the base class return `_shearType` if you access `shearType` (in which case, my comment about `NotImplementedError` is invalid). This is not any less read-only way to go about it.

Added!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -308,18 +324,10 @@ def fail(self, record, error=None):
...
-        if self.shearType != "BJ":
-            raise pexConfig.FieldValidationError(
-                self.__class__.shearType,
-                self,
-                f"shearType '{self.shearType}' is not valid. It must be 'BJ'.",
-            )
-        super().validate()
+    @HsmShapeConfig.shearType.getter
+    def shearType(self):
+        # Docstring inherited.
```
arunkannawadi, line 329: I am not so sure that the docstring will be inherited for property.  Running `help(HsmShapeBjConfig.shearType)` gets the docstring, but `HsmShapeBjConfig.shearType.__doc__` does not. And I don't know how `sphinx` will try to get it, and we can take it on in a different issue.

arunkannawadi (approved):
This is looking much better. Thanks Erfan!

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: Opened a new ticket for this: https://rubinobs.atlassian.net/browse/DM-45784

PR #75, opened 2024-08-08
DM-45519: Add importable HSM configuration function

parejkoj:


arunkannawadi:
This looks eerily similar to https://github.com/lsst/meas_extensions_shapeHSM/blob/main/config/enable.py . Any chance you can use that, or make this function use that instead of enabling the plugins in two different files?

parejkoj:
The goal will be to remove `config/enable`, because that's not importable, and thus can't really be used in a Task's `setDefaults`. See here for a use of it: https://github.com/lsst/drp_tasks/pull/67/files#diff-a7d4859560abf2b5f2beb220db84c1919101acb324d698dc4a98352fab6742a3R168

It doesn't look identical to `enable`, because I took the configuration that was in all of our obs-package overrides, since that's what we'll want as the default config.

arunkannawadi:
As long as `enable.py` would be removed either in this ticket, or soon enough, I am okay with it.

arunkannawadi:
Oh, also requesting to call this module `_configure`.

erykoff (approved):
Can this be used in more places?

parejkoj:
Probably, yes (coadd measurement in particular, I think), although many of those would probably go away once we remove `calibrate` from the DRP pipeline. I think I'd rather do the work to find those other uses on another ticket, though.

PR #77, opened 2024-09-09
DM-45784: Examine Sphinx docstring inheritance for overridden properties

enourbakhsh:


python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -41,6 +41,17 @@
 ]
 
 
+def inherit_prop_docs(cls):
+    """Decorator to inherit docstrings for properties from base classes."""
+    for name in dir(cls):
+        attr = getattr(cls, name, None)
+        if isinstance(attr, property):
+            base_prop = getattr(cls.__bases__[0], name, None)
+            if isinstance(base_prop, property) and attr.__doc__ is None:
```
arunkannawadi: Do we need a second `isinstance` check here?

arunkannawadi (commented):
The decorator seems a bit of an overkill here. There seem to be two property attributes: `history` (we don't care) and `shearType` and `dir(cls)` has 59 or so elements. Having these many `isinstance` checks to set one property's docstring, especially when this kind of a pattern is not so common, seems wasteful. I'd rather just add a line at the bottom of this file

```python
HsmShapeBjConfig.shearType.__doc__ = HsmShapeConfig.shearType.__doc__
```
and so on for each of the children classes. Given that we don't expect any new shape algorithms coming with HSM and any new property added, readability wins over generalizability to me.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
enourbakhsh: I agree there are too many `isinstance` checks. I used a different approach in my last commit.

enourbakhsh:
> The decorator seems a bit of an overkill here. There seem to be two property attributes: `history` (we don't care) and `shearType` and `dir(cls)` has 59 or so elements. Having these many `isinstance` checks to set one property's docstring, especially when this kind of a pattern is not so common, seems wasteful. I'd rather just add a line at the bottom of this file
> 
> ```python
> HsmShapeBjConfig.shearType.__doc__ = HsmShapeConfig.shearType.__doc__
> ```
> 
> and so on for each of the children classes. Given that we don't expect any new shape algorithms coming with HSM and any new property added, readability wins over generalizability to me.

Thanks, @arunkannawadi! Based on your feedback, I switched to a much simpler and shorter approach by placing the decorator on the property instead of the whole class. This avoids all the extra checks you mentioned in your comment.

arunkannawadi (approved):
This is looking so much better. Just confirm that the sphinx runs fine by building the docs for this package locally.

https://developer.lsst.io/stack/building-single-package-docs.html

PR #79, opened 2025-02-11
DM-42047: Update shapeHSM when optimized public galsim interface is available

enourbakhsh:


python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -122,33 +122,24 @@ def _calculate(
             Raised for errors in measurement.
         """
         # Convert centroid to GalSim's PositionD type.
+        # NOTE: Using _PositionD skips some sanity checks and argument parsing
```
arunkannawadi:
```suggestion
        # Using _PositionD skips some sanity checks and argument parsing
```

arunkannawadi: Also worth adding that `galsim` considers `_PositionD` as a public constructor despite the leading underscore.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -532,8 +522,6 @@ def _adjustNoise(
         record : `~lsst.afw.table.SourceRecord`
             The source record where measurement outputs will be stored. May be
             modified in place to set flags.
-        bounds : `~galsim.bounds.BoundsI`
-            The bounding box of the PSF as a GalSim bounds object.
```
arunkannawadi, line 536: Wow, I wonder what we intended this for.

arunkannawadi (approved):

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: The `bounds` parameter was needed for `badpix` but became unnecessary after changes in [this PR](https://github.com/lsst/meas_extensions_shapeHSM/pull/48/files). It was mistakenly left as an argument in `_adjustNoise()`.

python/lsst/meas/extensions/shapeHSM/_hsm_shape.py
```diff
@@ -262,44 +262,30 @@ def measure(self, record, exposure):
...
+                guess_sig_PSF=psfSigma,
+                precision=precision,
+                guess_centroid=guessCentroid,
+                strict=True,  # Raises GalSimHSMError if estimation fails.
+                check=False,  # This speeds up the code!
+                hsmparams=None,
             )
-        # GalSim does not raise custom pybind errors as of v2.5, resulting in
-        # all GalSim C++ errors being RuntimeErrors.
-        except (galsim.hsm.GalSimHSMError, RuntimeError) as error:
```
enourbakhsh, line 302: Also, I believe we no longer need to include `RuntimeError` here, as the exceptions are now raised by Python. I made the same change for `_hsm_moments`.

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
enourbakhsh: Done!

PR #82, opened 2025-04-21
DM-45569: Add color dependency when calling PSF.

PFLeget:


python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -44,6 +45,19 @@
...
+        c = record[columnNameValue]
+        ct = record[columnNameType]
+        if np.isfinite(c):
+            color = Color(colorValue=c, colorType=ct)
+        else:
+            color = Color()
+    except Exception:
+        color = Color()
+    return color
+
```
arunkannawadi: Let's actually move this more generally to `meas_base` and call it `colorExtractor`, similar to `centroidExtractor`. I am sure more algorithms could use this.

arunkannawadi (approved):

python/lsst/meas/extensions/shapeHSM/_hsm_moments.py
```diff
@@ -44,6 +45,19 @@
 ]
 
 
+def getPsfColorInfo(record, columnNameValue="psfColorValue", columnNameType="psfColorType"):
```
PFLeget: use snake_case

